PR #166, opened 2019-06-04
Fix160

### LINKED ISSUE #160: Python and node call, follow PyCall's example
To avoi issues with people not having `python3` or `node` as a way to call the relevant executable, it would be better to follow the route explained in https://github.com/JuliaPy/PyCall.jl where one can specify  `ENV` and change whatever path they want.

See also comment in  #159 

In windows' powershell, it looks like `py -3` is the right way to call things (see https://stackoverflow.com/questions/20364207/how-do-i-add-python-3-3-to-powershell) so there may need to be a system test etc. Maybe specifying ENV is the easier route.

tlienart:
Should eventually close #160 , needs testing on windows first.


tlienart:
This seems to work on windows.

PR #168, opened 2019-06-05
closes #162 allows a pre-path to be specified in the optimisation/pub…

### LINKED ISSUE #162: Paths specification for project site
As indicated in https://github.com/tlienart/JuDoc.jl/pull/159#issuecomment-498039034, the current path system works well for a situation where the folder maps to a `url.something/index` for instance 

```
username.github.io/index.html
```

and so forth. However one may want to do this for a relative path such as 

```
username.github.io/project/index.html
```

in which case all paths should be prepended with `/project/`. 

This should probably be specified in `config`. One thing to pay attention to is to keep the local version working fine too.

tlienart:
Following calls are allowed:

* `optimize(...; prepath="testproj")`
* `publish(...;prepath="testproj")`
* in config.md `@def prepath = "testproj"`

All this will result in the same thing (provided `optimize` and/or `publish` is used at some point): all paths `href="/..."` and `src="/..."` will be prepended with `/testproj/...`. For instance: https://tlienart.github.io/testproj/

This may not be the most elegant approach but it solves #162, we may revisit this in the future to do something that's more robust.

* edit: could do with tests and docs

### DIFF
--- a/src/converter/html.jl
+++ b/src/converter/html.jl
@@ -3,7 +3,8 @@ $(SIGNATURES)
 
 Convert a judoc html string into a html string (i.e. replace `{{ ... }}` blocks).
 """
-function convert_html(hs::AbstractString, allvars::JD_VAR_TYPE, fpath::AbstractString="")::String
+function convert_html(hs::AbstractString, allvars::JD_VAR_TYPE, fpath::AbstractString="";
+                      isoptim::Bool=false)::String
     # Tokenize
     tokens = find_tokens(hs, HTML_TOKENS, HTML_1C_TOKENS)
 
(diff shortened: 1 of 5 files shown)


danmrc:
Small issue: I tried `optimize(...; prepath="testproj")` and it throws:

> ERROR: MethodError: Cannot `convert` an object of type String to an object of type Pair{K,Tuple{Vararg{DataType,N}}} where N where K
Closest candidates are:
  convert(::Type{T}, ::T) where T at essentials.jl:154
Stacktrace:
 [1] setindex!(::Dict{String,Pair{K,Tuple{Vararg{DataType,N}}} where N where K}, ::String, ::String) at .\dict.jl:381
 [2] #optimize#127(::Bool, ::Bool, ::Bool, ::String, ::Function) at C:\Users\User\.julia\packages\JuDoc\4JWQH\src\manager\post_processing.jl:29
 [3] (::getfield(JuDoc, Symbol("#kw##optimize")))(::NamedTuple{(:prepath,),Tuple{String}}, ::typeof(optimize)) at .\none:0
 [4] top-level scope at none:0

Everything else is working! This include using the @def prepath and the `optimize()` command in windows

Thank you very much!

tlienart:
Thanks for reporting! the issue you reported is due to a typo on my side, it's fixed.

Glad that the rest works, don't hesitate to open other issues :)

Issue #174, opened 2019-06-12
Example not working on Windows


cserteGT3:
Edit: the basic template works, I think the problem is with my nodejs install, I will investigate it.
Sorry for the noise.
~I tried to run the example (as in the docs), but getting an ErrorException:~
```
Microsoft Windows [Version 10.0.17763.557]
(c) 2018 Microsoft Corporation. Minden jog fenntartva.
C:\Users\Ipse\Documents\Code>julia
               _
   _       _ _(_)_     |  Documentation: https://docs.julialang.org
  (_)     | (_) (_)    |
   _ _   _| |_  __ _   |  Type "?" for help, "]?" for Pkg help.
  | | | | | | |/ _` |  |
  | | |_| | | | (_| |  |  Version 1.1.0 (2019-01-21)
 _/ |\__'_|_|_|\__'_|  |  Official https://julialang.org/ release
|__/                   |

(v1.1) pkg> add JuDoc
  Updating registry at `C:\Users\Ipse\.julia\registries\General`
  Updating git-repo `https://github.com/JuliaRegistries/General.git`
 Resolving package versions...
 Installed LiveServer ───── v0.1.1
 Installed JuDoc ────────── v0.1.3
 Installed JuDocTemplates ─ v0.2.1
  Updating `C:\Users\Ipse\.julia\environments\v1.1\Project.toml`
  [4ca9428c] + JuDoc v0.1.3
  Updating `C:\Users\Ipse\.julia\environments\v1.1\Manifest.toml`
  [4ca9428c] + JuDoc v0.1.3
  [6793090a] + JuDocTemplates v0.2.1
  [16fef848] + LiveServer v0.1.1

julia> using JuDoc
[ Info: Recompiling stale cache file C:\Users\Ipse\.julia\compiled\v1.1\JuDoc\RFfo3.ji for JuDoc [4ca9428c-4c75-11e9-2efb-bf5cb6c1e8f8]
✘ Couldn't find highlight.js (`node -e "require('highlight.js')"` failed).
→ It is required for pre-rendering highlight.js but is not necessary to run JuDoc (cf docs).
✘ Couldn't find python3 (`python3 -V` failed).
→ It is required for minification but not necessary to run JuDoc (cf docs).
→ After installing any missing component, please re-build the package (cf docs).

julia> newsite("test", template="pure-sm")
✓ Website folder generated at "test" (now the current directory).
→ Use serve() from JuDoc to see the website in your browser.

julia> serve()
[...]

PR #178, opened 2019-06-16
Update highlight install command

cserteGT3:
I've experienced some issues with the `highlight.js` package installation, and finally this command solved it:
```bash
[sudo] npm install --save -g highlight.js
```
So `--save` is needed (in my case at least). I'm not experienced with node.js at all, so it's possible that it's not a perfect solution.
Sources:
[stackoverflow](https://stackoverflow.com/a/30886703), [meteor issue](https://github.com/meteor/meteor/issues/8119#issuecomment-264224373).

Edit: I think this caused #174 (at least the node.js part).

### DIFF
--- a/docs/src/index.md
+++ b/docs/src/index.md
@@ -34,7 +34,7 @@ This step requires a few external dependencies:
 Assuming you have those, you will then need to install `highlight.js` via `npm`:
 
 ```bash
-[sudo] npm install -g highlight.js
+[sudo] npm install --save -g highlight.js
 ```
 
 and the python package [`css_html_js_minify`](https://github.com/juancarlospaco/css-html-js-minify) which you can install with `pip3` (if you have python3, JuDoc will try to do this for you):


tlienart:
Thanks a lot for this!

mbaz:
I had the opposite problem:

```
$ sudo npm install --save -g highlight.js
+ highlight.js@9.15.8
updated 1 package in 0.384s

$node -e "require('highlight.js')"
internal/modules/cjs/loader.js:670
    throw err;
    ^

Error: Cannot find module 'highlight.js'
[...]

$ sudo npm install -g highlight.js
+ highlight.js@9.15.8
updated 1 package in 0.428s
$ node -e "require('highlight.js')"
$
```

cserteGT3:
To be precise, when I tried this (`--save`) on another Windows machine, I couldn't get it to work for first try. It worked only after running this and the original command couple of times, so I'm not sure that this is the way to do the installation. (On my laptop, this worked for the first time).
Maybe we should roll back to the original command and mention the linked posts? Then everyone can figure out if encounters the error.

(Btw. I'm thinking on switching to linux. I'm starting to hate windows :D )

tlienart:
Hmm shoot, I also did it like @mbaz and it worked.

I guess in want of a clear explanation, I'll change the doc slightly to say that the former should work, and that the `--save` may help otherwise e.g. on windows.

PR #183, opened 2019-06-20
fix typo by textoutput

cserteGT3:
I assume that this is a typo.

(Should I PR every small thing like this? Or just open an issue?)


tlienart:
Thanks!  PR or issue is fine either way, and very useful

PR #191, opened 2019-07-11
Add commitmsg kwarg for publish()

cserteGT3:
I think we should have this.

### DIFF
--- a/src/manager/post_processing.jl
+++ b/src/manager/post_processing.jl
@@ -75,7 +75,7 @@ In other scenarios you should probably do this manually.
 * `nopass=false`:   set this to true if you have already run `optimize` manually.
 """
 function publish(; prerender::Bool=true, minify::Bool=true, nopass::Bool=false,
-                   prepath::String="")::Nothing
+                   prepath::String="", commitmsg::String="jd-update")::Nothing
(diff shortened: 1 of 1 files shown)


src/manager/post_processing.jl
```diff
@@ -75,7 +75,7 @@ In other scenarios you should probably do this manually.
 * `nopass=false`:   set this to true if you have already run `optimize` manually.
 """
 function publish(; prerender::Bool=true, minify::Bool=true, nopass::Bool=false,
-                   prepath::String="")::Nothing
+                   prepath::String="", commitmsg::String="jd-update")::Nothing
```
tlienart: that's great, I would be in favour of just calling it `message` instead of `commitmsg`.

src/manager/post_processing.jl
```diff
@@ -75,7 +75,7 @@ In other scenarios you should probably do this manually.
 * `nopass=false`:   set this to true if you have already run `optimize` manually.
```
tlienart: Could you add a line here for the new kwarg?

cserteGT3: Sorry, I missed that. Added below.

PR #196, opened 2019-08-07
resolve highlight.js installation instruction 

cserteGT3:
Again... I tried to set up `highlight.js` and the proposed advice (#178) did not work, so searched a bit more. [What I've found](https://stackoverflow.com/a/26480275) is the following: putting the `node_modules` folder in an environment variable. I tested it with removing-reinstalling (and trying with other suggestions too) and this one worked.
If this works for @mbaz too, that would be a good confirmation.

### DIFF
--- a/docs/src/index.md
+++ b/docs/src/index.md
@@ -47,9 +47,6 @@ Assuming you have those, you will then need to install `highlight.js` via `npm`:
 [sudo] npm install -g highlight.js
 ```
 
-On Windows you _may_ need to add the following to your environment variables ([source](https://stackoverflow.com/a/26480275)):
-`NODE_PATH=%AppData%\npm\node_modules`
-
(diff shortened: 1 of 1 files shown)


docs/src/index.md
```diff
@@ -47,7 +47,8 @@ Assuming you have those, you will then need to install `highlight.js` via `npm`:
 [sudo] npm install -g highlight.js
 ```
 
-(you _may_ need to add `--save` to make this work well, see if the tests below work and if not, try again with `--save`)
```
tlienart: Maybe don't remove this entirely and wrap something together with "if the tests don't pass, you can try this (bullet point) or this (bullet point)". I don't understand enough about node to see if there's one fool proof approach so probably best to leave all things that may help here and wait for more feedback from other users.

cserteGT3: Me neithert. I'll do that.

cserteGT3:
I've organized the suggestions into a section (below Troubleshooting). Let me know what you think about it.

tlienart:
thanks!

Issue #195, opened 2019-08-07
Inserting tables [WIP]


cserteGT3:
> ### PS
> 
> When I was initially working on these insertion commands, I thought it could be useful to have something for automatic table insertions, for instance if a table was generated by a script and then it would be formatted nicely in markdown. It sounds like this may be what you're trying to do. If so and if you'd be interested in this, maybe open another issue and we can work through it? (I reckon it would be a simple PR if you're feeling adventurous)
> 
> _Originally posted by @tlienart in https://github.com/tlienart/JuDoc.jl/issues/193#issuecomment-518593880_

I'm curious about your idea and happy to contribute to your work!

tlienart:
Nice!

Well I guess there could be several ways of doing this but the idea was:

1. user has some CSV file, possibly with headers, (e.g. CSV file was generated)
2. there's a command  `\tableinput{}{...}` which reads such a CSV file and inserts formatted markdown.

The first `{}` could correspond to a header specification e.g. `col1, col2, col3` or empty if to be deduced from the file. 
The second `{}` would contain the path to the file. 

For the command, you could mimick the function `resolve_lx_figalt` in `src/converter/lx_simple.jl` and define a simpler `resolve_lx_tableinput`. 

Then you need to add it to the dictionary `LXCOM_SIMPLE` in the same file and indicate that it's a default command in `def_GLOBAL_LXDEFS!` in `src/jd_vars.jl`

(I reckon if you just grep for `figalt` and mimick the behaviour adapting for a table input instead of a figure, it should more or less guide you).

cserteGT3:
Here's couple of my thoughts (after diving into the package's code):
1. We should decide, if we want `CSV.jl` as dependency. It's doable with `readdlm()`, but I guess it will be more restrictive (one example: the header is decided by the user OR from the CSV file, the user can not override the header in the file). For me it's fine either way, though `readdlm()` would be more work (which is also more fun I guess 😄 ). Also I didn't have an MWE to state these things.

2. `resolve_lx_tableinput` should generate what? Html or markdown that is then passed to another function (which I didn't find yet) which parses it to html?

3. Would be nice to have a 3rd argument where the user can specify the alignment (after we've got a working version of the basic functionality).

tlienart:
[...]

PR #197, opened 2019-08-11
WIP: inserting tables with \tableinput command

cserteGT3:
As discussed in #195 this PR adds the `\tableinput{header}{path}` command. The current implementation works in the sense that the markdown strings are generated and included in the page. As I see, the followings are need to be done:
- [x] conversion to html
- [x] error handling: check if the provided header's length==number of columns
- [x] tests
- [x] docs

I need help with the first two tasks (and will do the other two later on):
1. I just don't know which function to call on the result string to convert it to html.
2. If the header's and the table's length doesn't match: error should be thrown or an `@assert` is fine, or should just ignore it and write an error message?


src/converter/lx_simple.jl
```diff
@@ -154,6 +176,7 @@ const LXCOM_SIMPLE = Dict{String, Function}(
     "\\output" => resolve_lx_output,  # include plain output generated by code
     "\\figalt" => resolve_lx_figalt,  # include a figure (may or may not have been generated)
     "\\file"   => resolve_lx_file,    # include a file
+    "\\tableinput" => resolve_lx_tableinput,    # include table from a csv file
```
tlienart: you want this to be in LXCOM_SIMPLE_REPROCESS; the reprocess bit just means that the markdown will be re-parsed by JuDoc.

So for instance `\\figalt` directly includes HTML so it does not need to be reprocessed but `\\tableinput` returns MD which needs to be reprocessed.

tlienart:
Ok I think your solution is a bit too complicated:

Let's say that the file is `test.csv`

```
a, b, c, d
0, 1, 2, 3
4, 5, 6, 7
```

Then 

```julia
using DelimitedFiles
content = readdlm("test.csv", ',', String)
open("out.md", "w") do fout
    for row in eachrow(content)
        write(fout, "| " * prod(e * " | " for e in row))
        write(fout, '\n')
    end
end
```

Does most of the job (header part missing but it's easy to add). 

For your question (1) see review comment, just declare the command in `LX_SIMPLE_REPROCESS`, nothing else.

For (2), an `html_err` is probably best so that it doesn't stop the engine but just writes a message on the HTML page.

cserteGT3:
Only reason for this ([from the docs](https://docs.julialang.org/en/v1/stdlib/Markdown/#Tables-1)):
 > As illustrated in the above example each column of `|` characters must be aligned vertically.

If the cells in a column doesn't have the same length (like in the followings), `|` will not be aligned:
```
a, b, c, d
string1, 1.123, 2, 3
string126, 0.1, 6, 7
```
Result markdown:
```
| a |  b |  c |  d | 
| string1 |  1.123 |  2 |  3 | 
| string126 |  0.1 |  6 |  7 | 
```
I didn't tested how the parser handles this kind of table, but in the docs it's stated that they must align, so I would rather align them.

tlienart:
Here's a bit more code which may help with testing etc: 

```julia
using Markdown

test = """
    This is a table:

    | a |  b |  c |  d |
    | :---: | :---: | :---: | :---: |
    | string1 |  1.123 |  2 |  3 |
    | string126 |  0.1 |  6 |  7 |

    Done.
    """

result = test |> Markdown.parse |> Markdown.html
write("index.html", result)
```

if you open that in a browser you'll get:

Screen Shot 2019-08-13 at 09 47 13

which is what we want (separators etc are styled in CSS).

By the way, JuDoc also uses this `Markdown.parse |> Markdown.html` chain though after a bit of processing.

cserteGT3:
I'll try to answer everything in this comment.
My two questions:
As discussed `html_err()` is used to handle `DimensionMismatch` of the header and if the file is not present. Therefore I left the command in `LX_SIMPLE` and used the `md2html` function to convert the generated markdown to html. Hope that is a good solution.

I experienced too, that alignment doesn't matter for the parser, moreover if `|` are aligned, the parser can't parse the string into the table, which I don't really understand. I may open an issue, because currently the behaviour is totally different than the documentation 😄 

I'll soon figure out the testing, and write the docs.

tlienart:
Hey thanks a lot for this, I'll try to look at this tomorrow and get back to you with, hopefully helpful, comments.

PS: given that you're one of the first to actually have a look at the code, don't hesitate to also let me know the things that don't make sense and I'll try to add comments (or docs) as appropriate to guide future contributions

cserteGT3:
A quick note:
While I was writing the docs I realized that `Documenter` needs to have the `|`-s aligned.

cserteGT3:
I've finished the first version of documentation and tests. Let me know if they serve their purpose.

PS: This is my first larger PR, so I'm not sure what else should be done (after approving the changes). Rebasing and/or squashing commits? If so, please do let me know, I'd like to do them to practice open source development 😃  .

cserteGT3:
Answering to my own issue:
> 
> I experienced too, that alignment doesn't matter for the parser, moreover if `|` are aligned, the parser can't parse the string into the table, which I don't really understand. I may open an issue, because currently the behaviour is totally different than the documentation 😄

I started writing an MWE and realized that the parser works totally fine. Also I misunderstood what you've written. Now I understand that CSS is dominant.

docs/src/man/syntax.md
```diff
@@ -681,6 +681,51 @@ This is the index then some **markdown** in a side file.
 
 **Note**: if you don't specify a file extension, `.md` is appended to the specified path.
 
+### Inserting a table
+
+You can insert tables directly from CSV files with the `\tableinput{header}{path}` command.
+If you generate the file on-the-fly, you should follow this example:
+`````judoc
+```julia:./tableinput/gen
+testcsv = "h1,h2,h3\n152,some string, 1.5f0\n0,another string,2.87"
```
tlienart: write

```
testcsv = """
h1,h2,h3
152, some string, 1.5f0
0, another string, 2.87
"""
```
for readability

docs/src/man/syntax.md
```diff
@@ -681,6 +681,51 @@ This is the index then some **markdown** in a side file.
...
+`````judoc
+\tableinput{}{./tableinput/testcsv.csv}
+`````
+Which will result in:
+
+| h1  | h2             | h3    |
+| --- | -------------- | ----- |
+| 152 | some string    | 1.5f0 |
+| 0   | another string | 2.87  |
+
```
tlienart: Add something like `in this case given no header was specified in the call, a header was generated from the first line in the CSV, here  h1, h2, h3`

docs/src/man/syntax.md
```diff
@@ -681,6 +681,51 @@ This is the index then some **markdown** in a side file.
...
+`````judoc
+\tableinput{}{/assets/test2.csv}
+`````
+
+!!! note
+
+     The look of the table will be defined by your CSS stylesheet.
+
+There's a couple of rules that you have to keep in mind when using the `\tableinput{}{}` command:
+* The CSV file must have a header line.
```
tlienart: ah right so at the moment you only allow empty header argument?

docs/src/man/syntax.md
```diff
@@ -681,6 +681,51 @@ This is the index then some **markdown** in a side file.
...
+* Columns must be separated with comma (`,`).
+* If header is specified, its length must be matched with the table in the file.
```
tlienart: If _a_ header ... must match the number of columns of the file

src/converter/lx_simple.jl
```diff
@@ -68,6 +68,70 @@ function resolve_lx_figalt(lxc::LxCom)::String
...
+
+"""
+$SIGNATURES
+
+Internal function to process an array of strings to markdown table (in one single string).
+If header is empty, the first row of the file will be used for header.
+"""
+function csv2html(path, header)::String
+    csvcontent = readdlm(path, ',', String, header=false)
+    tablesize = size(csvcontent)
```
tlienart: I think it would be more readable to write

```
nrows, ncols = size(csvcontent)
```

src/converter/lx_simple.jl
```diff
@@ -68,6 +68,70 @@ function resolve_lx_figalt(lxc::LxCom)::String
...
+    # replacing header
+    if ! isempty(header)
+        newheader = split(header, ",")
+        hs = size(newheader,1)
+        hs != tablesize[2] && return html_err("header ($hs) and table ($(tablesize[2])) size does not match")
```
tlienart: size <> number of columns

src/converter/lx_simple.jl
```diff
@@ -68,6 +68,70 @@ function resolve_lx_figalt(lxc::LxCom)::String
...
+        csvcontent[1,:] = newheader
+    end
+    io = IOBuffer()
+    # writing the header
+    for a in csvcontent[1,:]
```
tlienart: This is a bit of microoptimization probably but it's better to not create this copy `csvcontent[1, :]`, prefer something like

```
for c in 1:ncols
    write(io, "| ", csvcontent[1, i])
end
```

tlienart: probably this would work too; though probably not faster as writing to IOBuffer is quick.

```
write(io, prod("| " * csvcontent[1, i] for i in 1:ncols))
```

tlienart:
Ok, I've added some feedback on "just the code", will now pull your branch, play a bit with it and see if it works as expected, assuming that  + fixing of minor comments above, I'll merge. Thanks for contributing!

tlienart:
Ok it mostly looks good, one more thing though: this test:

```
# header specified
    h = raw"""
        A table:
        \tableinput{A,B,C}{/assets/testcsv.csv}
        Done.
        """ |> seval
    shouldbe = """<p>A table: <table><tr><th>A</th><th>B</th><th>C</th></tr>
            <tr><td>string1</td><td>1.567</td><td>0</td></tr>
            <tr><td></td><td></td><td></td></tr>
            <tr><td>l i n e</td><td>.158</td><td>99999999</td></tr></table>
            Done.</p>"""
    @test isapproxstr(h, shouldbe)
```

looks incorrect; when the header is specified you should not over-write the first line of the CSV file, so the first row after A, B, C should be h1, h2, h3  which is present in the file

docs/src/man/syntax.md
cserteGT3: I'm not sure I understand what you mean. Should it be here `\tableinput{header}{path}`?

src/converter/lx_simple.jl
cserteGT3: Should the error message be something like:
`html_err("header size ($hs) and number of cloumns ($(tablesize[2])) does not match")`
?

docs/src/man/syntax.md
tlienart: I was a bit confused when reading this and not having seen the code, it seems to imply that:

1. the CSV file **must** have a header file which
2. you will use as the header of the table

this would then imply that you're never using the first element i.e. that if you use `\tableinput{h1, h2, h3}{...}` it would ignore the `{h1, h2, h3}`. I'm assuming now from having read the rest of your code that it's not true, so the phrasing here should change to indicate how to use it and also remove the part about the CSV that **must** have a header (not the case if you provide one). Hope that makes sense

src/converter/lx_simple.jl
tlienart: yeah, a few fixes for the english:

- cloumns , typo -> columns
- does not -> do not

src/converter/lx_simple.jl
cserteGT3: A quick benchmark just for fun:
The three approaches:
1. `csv2md_old()` - original
1. `csv2md_old2()` - `for c in 1:ncols` instead of copying
1. `csv2md_new()` - using `prod()`

## Code

A bit modified, html conversion is removed.
```julia
function csv2md_old(path, header)::String
    csvcontent = readdlm(path, ',', String, header=false)
    nrows, ncols = size(csvcontent)
    # replacing header
    if ! isempty(header)
        newheader = split(header, ",")
        hs = size(newheader,1)
        hs != ncols && return "header ($hs) and table ($ncols) size does not match"
        csvcontent[1,:] = newheader
    end
    io = IOBuffer()
    # writing the header
    for a in csvcontent[1,:]
        write(io, "| ", a, " ")
    end
    # writing end of header & header separator
    write(io, "|\n|", repeat( " ----- |", ncols), "\n")
    # writing content
    for i in 2:nrows
        for j in 1:ncols
            write(io, "| ", csvcontent[i,j], " ")
        end
        write(io, "|\n")
    end
    return String(take!(io))
end

function csv2md_old2(path, header)::String
    csvcontent = readdlm(path, ',', String, header=false)
    nrows, ncols = size(csvcontent)
    # replacing header
    if ! isempty(header)
        newheader = split(header, ",")
        hs = size(newheader,1)
        hs != ncols && return "header ($hs) and table ($ncols) size does not match"
        csvcontent[1,:] = newheader
    end
    io = IOBuffer()
    # writing the header
    for i in 1:ncols
        write(io, "| ", csvcontent[1,i], " ")
    end
    # writing end of header & header separator
    write(io, "|\n|", repeat( " ----- |", ncols), "\n")
    # writing content
    for i in 2:nrows
        for j in 1:ncols
            write(io, "| ", csvcontent[i,j], " ")
        end
        write(io, "|\n")
    end
    return String(take!(io))
end

function csv2md_new(path, header)::String
    csvcontent = readdlm(path, ',', String, header=false)
    nrows, ncols = size(csvcontent)
    # replacing header
    if ! isempty(header)
        newheader = split(header, ",")
        hs = size(newheader,1)
        hs != ncols && return "header ($hs) and table ($ncols) size does not match"
        csvcontent[1,:] = newheader
    end
    io = IOBuffer()
    # writing the header
    write(io, prod("| " * csvcontent[1, i] * " " for i in 1:ncols))
    # writing end of header & header separator
    write(io, "|\n|", repeat( " ----- |", ncols), "\n")
    # writing content
    for i in 2:nrows
        for j in 1:ncols
            write(io, "| ", csvcontent[i,j], " ")
        end
        write(io, "|\n")
    end
    return String(take!(io))
end
```
## Results

(CSV file is around 200 rows and of course they produce the same string)
```julia
julia> csv2md_new("tcsv.csv","") == csv2md_old2("tcsv.csv","") == csv2md_old("tcsv.csv","")
true

julia> @benchmark csv2md_old($"tcsv.csv",$"")
BenchmarkTools.Trial:
  memory estimate:  84.98 KiB
  allocs estimate:  1396
  --------------
  minimum time:     383.200 μs (0.00% GC)
  median time:      448.800 μs (0.00% GC)
  mean time:        488.064 μs (3.76% GC)
  maximum time:     54.816 ms (99.02% GC)
  --------------
  samples:          10000
  evals/sample:     1

julia> @benchmark csv2md_old2($"tcsv.csv",$"")
BenchmarkTools.Trial:
  memory estimate:  84.86 KiB
  allocs estimate:  1394
  --------------
  minimum time:     368.900 μs (0.00% GC)
  median time:      423.199 μs (0.00% GC)
  mean time:        456.610 μs (3.33% GC)
  maximum time:     55.527 ms (99.08% GC)
  --------------
  samples:          10000
  evals/sample:     1

julia> @benchmark csv2md_new($"tcsv.csv",$"")
BenchmarkTools.Trial:
  memory estimate:  85.13 KiB
  allocs estimate:  1403
  --------------
  minimum time:     375.300 μs (0.00% GC)
  median time:      429.701 μs (0.00% GC)
  mean time:        469.463 μs (3.29% GC)
  maximum time:     55.676 ms (99.06% GC)
  --------------
  samples:          10000
  evals/sample:     1
```
I think we should go with the one line version, I like that I learned something new (didn't know about `prod()`.

tlienart: nice ;-)

src/converter/lx_simple.jl
cserteGT3: Thank you! I'll fix it.

docs/src/man/syntax.md
cserteGT3: The CSV file must have a header line, even if it not used. I had a reason for doing this way, but as I checked now the code we can change this if this is confusing.
I'll definitely change the phrasing (after we decide on the behaviour).

tlienart: Ok I don't understand this, what I thought was:

* case 1: no header provided; use the first line of the CSV as header
* case 2: header provided; act as in case 1 but after effectively adding the header provided on top as a first row

am I missing something?

cserteGT3: That is correct, currently this happens. Should we change this?

tlienart: see my comment in the test bit, it was not happening (you may have changed that now) where if a header was provided you **replaced** the first line. 

In any case if you agree with my previous comment, then this issue was mostly about the phrasing in the docs so that it becomes clear how to provide a header and what happens when you do 😄

cserteGT3: Sorry, my bad. I misunderstood again. Now the following happens:
* case 2: header provided `\tableinput{header}{path}`; the provided header will be used instead of the header in the CSV.

So currently must be a header in the CSV, which you can override with `\tableinput{header}{path}`.

tlienart: Ok no, I don't think that's what we want; imagine the following situation:

1. user generates array of numbers, three columns
2. user wants to display numbers with name of columns `x1, x2, x3`
3. user calls `\tableinput{x1, x2, x3}{path}`

if at (3) the user doesns't see the first line of the table, the user will not be happy.

tlienart: that's why I said; effectively case 2 should be considered as if the user provided an extra line which is slapped on top of the content of the CSV, and then you apply case 1 where the first line is read as a header.

cserteGT3: The workflow in my mind was, that you have a CSV and you may want to display an other header line, than in the file (nicer/better name, etc.).
Thinking on it a bit, you're solution is more generic as it can be used for CSVs with or without header.

cserteGT3: I've pushed most changed, except this one. But I think this should be changed too.

docs/src/man/syntax.md
```diff
@@ -701,8 +703,8 @@ Which will result in:
 | --- | -------------- | ----- |
 | 152 | some string    | 1.5f0 |
 | 0   | another string | 2.87  |
-
-or if you want to specify the headers:
+In this case given no header was specified in the call, a header was generated from the first line in the CSV (here: h1, h2).
```
tlienart: here: `h1, h2, h3`)

test/converter/lx_simple.jl
```diff
@@ -49,3 +49,41 @@ end
...
+            Done.</p>"""
+    @test isapproxstr(h, shouldbe)
+    # header specified
+    h = raw"""
+        A table:
+        \tableinput{A,B,C}{/assets/testcsv.csv}
+        Done.
+        """ |> seval
+    shouldbe = """<p>A table: <table><tr><th>A</th><th>B</th><th>C</th></tr>
+            <tr><td>string1</td><td>1.567</td><td>0</td></tr>
```
tlienart: so yeah please change this test so that the given header is _added_ above the first line and does not replace the first line, once that's good, I'm good to merge, thanks!

cserteGT3:
Sorry for the silence, I had a busy week. I've updated the code, tests and documentation (and resolved every review). Please review them particularly the documentation

docs/src/man/syntax.md
```diff
@@ -681,6 +682,55 @@ This is the index then some **markdown** in a side file.
...
+`````
+Which will result in:
+
+| h1  | h2             | h3    |
+| --- | -------------- | ----- |
+| 152 | some string    | 1.5f0 |
+| 0   | another string | 2.87  |
+In this case given no header was specified in the call, a header was generated from the first line in the CSV (here: h1, h2, h3).
+
+If you're file doesn't have a header, you can specify it in the call:
```
tlienart: you're -> your

docs/src/man/syntax.md
```diff
@@ -681,6 +682,55 @@ This is the index then some **markdown** in a side file.
...
+| 0         | another string | 2.87      |
+
+With the above in mind, you can also include existing CSV files.
+
+!!! note
+
+     The look of the table will be defined by your CSS stylesheet.
+
+There's a couple of rules that you have to keep in mind when using the `\tableinput{}{}` command:
+* Columns must be separated with comma (`,`).
```
tlienart: with -> by a

docs/src/man/syntax.md
```diff
@@ -681,6 +682,55 @@ This is the index then some **markdown** in a side file.
...
+* If a header is specified, its length must match the number of columns of the file.
```
tlienart: of the file -> in the file

tlienart:
looks good to me minus minor nitpicking, thanks a lot for this Tamás! and actually given it's minor, I'll just merge and do it :-)

Issue #147, opened 2019-05-05
[v0.3.0] list


tlienart:
**docs**

* [x] document `\\`
* [x] document html entities and backslash escaping etc
* [x] document single, double and triple backticks 
* [x] document indirect links (images and href) + that this `[link text](https:://www.mozilla.org/ "Mozilla")` is not supported (where the `"Mozilla"` bit should be the link title).
  * [x] recommend not to use things that could clash with code so for instance, do not use `[link][1]` if there may be some place in your code with `[1]` otherwise both will be replaced and it'll be a mess. so either prefer inline links or use ids which don't clash.
* [x]  document that headers should only be defined via `#...` (also for TOC generation)
* [x] indented code blocks, why fenced is better (risk of clashes with lists, doesn't have the eval possibility and somewhat less battle tested)  but otherwise ok
* [x] document differences if come from other ssg (see #201 and the last comment)
* [x] document use of `lang` local page variable for indented code blocks

**think**

* [x] either brainstorm or document; if something like `` `[...]: blah` `` is found it will be replaced (as per indirect link processing) though it shouldn't be because it's part of a code block; maybe we can just assume that won't happen too much and could offer a toggle to only have inline links (so no delayed links) in which case this wouldn't happen.

**done**

* [x] #207 indented code block (via PR #217)
* [x] showing error message when an eval'd block fails (PR #187)
* [x] headers are now links
* [x] automatic table of contents via `\toc` or `\tableofcontents` (PR #188) 
  * [x] docs
* [x] major cleanup of the code (trying to make naming more consistent and less clunky, general cleaning up etc)
* [x] added `\\` as a way to force the introduction of a line skip, this can be useful in the context of inclusions etc (see https://github.com/cserteGT3/JuDocPlottest/issues/1 for context)
* [x] added fixes for backslashes in code environment etc (issue #205, PR #209)
* [x] html entities are now supported (issue #206, PR #209)
* [x] better status messages for `cleanpull` (PR #190) and adding the possibility to specify a commit message for `publish` (PR #191, creds to @cserteGT3) 
* [x] added `\textinput` command (PR  #194)
  * [x] docs
* [x] added `\tableinput` command (PR #197, creds to @cserteGT3) 
  * [x] docs
* [x] added support for double backticks (see issue #204 and PR #210)
* [x] added support for indirect links and images (PR #214) 
[...]

PR #217, opened 2019-09-09
closes #207 through introduction of LR_INDENT tokens

### LINKED ISSUE #207: Allow and/or automatically replace indented code blocks
In Jekyll and elsewhere it seems that people like the 4-space or 1-tab indented stuff like

```
    this is some
    code
```

which would appear as such. At the moment JuDoc only takes blocks that are fenced with `` ``` `` as code blocks (and this should be encouraged anyway to be compatible with the code execution thing)

tlienart:
* allows for indented code block
* respects common mark precedence (if indented line starts with something that looks like a list, the list will have precedence, to avoid clashes, use fenced code blocks)
* adds tests

last step will be to convert JuliaBlog posts _en masse_

### DIFF
--- a/src/converter/md.jl
+++ b/src/converter/md.jl
@@ -33,11 +33,14 @@ function convert_md(mds::String, pre_lxdefs::Vector{LxDef}=Vector{LxDef}();
     #> 1. Tokenize
     tokens = find_tokens(mds, MD_TOKENS, MD_1C_TOKENS)
 
+    #> 1'. Find indented blocks
+    tokens = find_indented_blocks(tokens, mds)
+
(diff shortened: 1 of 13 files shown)


cormullion:
I'm happy to have another pass through the set of Julia blog posts some time. Perhaps when the next release arrives...?

tlienart:
🥇 thanks! 🥇

The next release is very near as you might have seen in #147, I just want to update the docs with things that have been added/fixed since 0.2. 

I'll maybe try a few posts myself first (in particular a few that you've mentioned from Stefan and a few that contain maths) and then I'll ping you if that's ok with you?

PR #237, opened 2019-09-28
Fix nested table of contents HTML

TotalVerb:
HTML nested lists need to have the sublists [contained within the parent \<li\>](https://stackoverflow.com/questions/5899337/proper-way-to-make-html-nested-list). This PR changes the TOC generation to perform this nesting, as well as to detect the highest heading level (so as not to generate an empty H1 entry if the page contains no H1 headings).

Before this PR, note how the numbering is strange (because the \<ol\> tags are not contained within the list item, so essentially numbering is reset to 1 whenever an \<ol\> tag is seen, instead of nesting):

Screenshot from 2019-09-28 16-48-18

After this PR, the problem is fixed:

Screenshot from 2019-09-28 17-20-36

### DIFF
--- a/src/converter/html_functions.jl
+++ b/src/converter/html_functions.jl
@@ -102,33 +102,38 @@ The split is as follows:
 * h[4] indicates the level of the title
 """
 function hfun_toc()::String
-    inner  = ""
-    curlvl = 1
+    inner   = ""
+    baselvl = minimum(h[4] for h in values(PAGE_HEADERS)) - 1
+    curlvl  = baselvl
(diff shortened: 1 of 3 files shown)


tlienart:
Hello and thanks so much for this! I've had a brief look and it looks great to me; it's a bit late here so I'll have a more in depth look tomorrow and will merge. Thanks a lot for taking the time.

TotalVerb:
Of course, thank you so much for making this package!

tlienart:
lgtm, thanks!

PR #238, opened 2019-09-28
Fix search for occur to account for _n headings

TotalVerb:
Currently, only the first duplicated heading will have its ID renumbered, because the search for occur only looks for headings with the same ID. All of the duplicate headings after the first will be called `heading_2`:

```
### Heading (heading)
### Heading (heading_2)
### Heading (heading_2)
```

This changes the logic to look for all headings with the same refstring. There is probably still a problem if the user creates a document like

```
### Heading 2 (heading_2)
### Heading (heading)
### Heading (heading_2)
```

since the search won't find the explicit `heading_2` refstring. I'm not sure how to address this cleanly, however, so I opted for a more minimal change.

### DIFF
--- a/src/converter/md_blocks.jl
+++ b/src/converter/md_blocks.jl
@@ -107,7 +107,7 @@ function convert_header(β::OCBlock)::String
     # check if the header has appeared before
     rstitle  = refstring(title)
     level    = parse(Int, hk[2])
-    occur    = (hv[3] for hv ∈ values(PAGE_HEADERS) if hv[2] == rstitle)
+    occur    = (hv[3] for hv ∈ values(PAGE_HEADERS) if refstring(hv[1]) == rstitle)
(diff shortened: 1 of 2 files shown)


tlienart:
I guess one way to go is to do something a bit like in Documenter where the first occurence gets a `_1` and then `_2` etc etc so in the pathological situation you suggest:

```
### Heading 2 (heading_2_1)
### Heading (heading_1)
### Heading (heading_2)
```

TotalVerb:
Yeah, Documenter.jl always numbers its links. But making that change would break existing section links generated by JuDoc, and in my opinion makes the links less understandable in the common case.

Another option is to reserve some sequence like double underscore `__` for numbering, and make refstring collapse double underscore to single underscore, which would probably be a little less disruptive?

tlienart:
Sounds good. I'll merge this and open an issue with your suggestion for consideration, thanks!

PR #239, opened 2019-09-28
Fix crash if non-ASCII character precedes $ math

TotalVerb:
Previously, if a sequence like `“$x$”` occurs (where the quotes are fancy quotes), then there would be a `StringIndexError`. This fixes the problem by calling `prevind` instead of subtracting `1`.

Since `prevind` with the default `n=1` can't return a negative number, I've also removed the else branch here since it's only setting 0 to 0.

### DIFF
--- a/src/converter/js_prerender.jl
+++ b/src/converter/js_prerender.jl
@@ -97,7 +97,7 @@ function js2html(hs::String, jsbuffer::IOBuffer, matches::Vector{RegexMatch},
     head, c = 1, 1
     for i ∈ 1:2:length(matches)-1
         mo, mc = matches[i:i+1]
-        write(htmls, subs(hs, head, mo.offset - 1))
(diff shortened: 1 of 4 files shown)


TotalVerb:
The second commit fixes a related issue, but during the prerendering KaTeX step. I split up the prerendering test so that I could test it locally, since I haven't been able to figure out node/highlight.js yet.

TotalVerb:
The Travis 1.0 failure is real, but preexisting ([as redirecting stderr/stdout to a IOBuffer is a 1.1 feature](https://discourse.julialang.org/t/how-to-capture-stdout-and-stderr-in-1-1/20712)). I guess Travis wasn't testing this codepath because it's missing highlight.js; this change made the failure visible by removing that check for this branch.

Maybe upgrading the minimum Julia version to 1.0 is reasonable?

PR #279, opened 2019-10-21
update to literate script should trigger

tlienart:
This needs to be tested and valiated before it can be merged


tlienart:
Actually this is working fine :-)

Issue #294, opened 2019-11-14
Allow the use of a global installation of Node


trigaut:
Hi, first thanks for that great package!
I am developing a collaborative blog using JuDoc.

I may need to deploy JuDoc in a docker container with a Node.js application which has a list of dependencies in its package.json file.
I just need to run `npm install` in the application folder to install all the js dependencies. 
This package.json is the perfect place to put the JuDoc dependency highlight.js. 
However, since issue #262 and PR #263, JuDoc uses exclusively NodeJS.jl and its "julia specific" node installation, which is not convenient as I install highlight.js using my docker image "global" node install. 

Are you against something like the following in the build.jl of JuDoc for people who want to choose the node install?
```julia
const NODE = begin
    if "NODE" ∈ keys(ENV)
        ENV["NODE"]
    else
        NodeJS.nodejs_cmd()
    end
end
```
I tested it works.
It would be consistent with the doc troubleshooting section https://tlienart.github.io/JuDoc.jl/dev/#Troubleshooting-1.

Moreover it is related to NodeJS but the function, `NodeJS.nodejs_cmd`, to launch node,
errors in a docker with alpine 3.10.2 and julia1.2.0.

Building the Dockerfile:
```dockerfile
FROM fredrikekre/julia-alpine:1.2.0 
RUN julia -e 'import Pkg; Pkg.add("NodeJS")' 
RUN julia -e 'using NodeJS; run(NodeJS.nodejs_cmd())'
```
returns
```bash
Step 3/3 : RUN julia -e 'using NodeJS; run(NodeJS.nodejs_cmd())'
 ---> Running in 07e33df46554
ERROR: IOError: could not spawn `/root/.julia/packages/NodeJS/rx0mH/deps/bin/node-v10.15.0-linux-x64/bin/node`: no such file or directory (ENOENT)
```

tlienart:
Hey thanks a lot, I'll review your proposal asap, thanks for taking the time! (I'm keen to have an easy way for most people and a specific way  for folks like you who know their way around NPM 😋)

The project sounds great btw!

tlienart:
Ok so now with a bit more time to look at what you're suggesting, it seems perfectly reasonable, thanks a lot, would you mind opening a PR? otherwise I'll do it tomorrow.

As a side note, I think you should report the error on Docker to the NodeJS repo!

[...]

PR #295, opened 2019-11-15
Edit build.jl to use global Node installation if the user sets ENV["NODE"]

trigaut:
Follows the suggestion of Issue #294


tlienart:
Thanks!

PR #297, opened 2019-11-20
Update NodeJS bounds

davidanthoff:
I released a new version of NodeJS.jl that updates the version of node used to v12.13.0. Ideally you would test whether your package works with that newer version of node, and then merge this here. You might also want to adjust the bounds in the general registry if you think that older versions of your package are also functional with the new release of NodeJS.jl.


tlienart:
Thanks  a  lot for the heads up! I've  tested this locally and it works fine. I've merged your branch in #298 so will close the PR here but thanks again for it!

PR #309, opened 2019-12-10
Feature request : callback after built page

MarcMush:
this allows to define a function that will be executed each time a page is built

in my case, I wanted to save all the jd_vars in an external file for later use, so I will overload the function  `on_page_built` with : 
```julia
function JuDoc.on_page_built(jd_vars)
    d = Dict()
    for (key, value) in pairs(jd_vars)
        d[key] = value |> first
    end

    output = joinpath("assets/vars", JuDoc.change_ext(basename(d["jd_rpath"]), ".json"))
    open(output, "w") do f
        write(f, JSON.json(d, 2))
    end
end
```

I'm not sure if it's the right way to do it, but this way it doesn't change the current behaviour or built time when this function is unchanged, but allows for more complex behaviours

### DIFF
--- a/src/manager/file_utils.jl
+++ b/src/manager/file_utils.jl
@@ -20,6 +20,14 @@ end
 """
 $(SIGNATURES)
 
+Executed after each page is rendered. Can be overwritten.
+"""
+function on_page_built(jd_vars) end
+
+
+"""
+$(SIGNATURES)
+
 Take a path to an input markdown file (via `root` and `file`), then construct the appropriate HTML
 page (inserting `head`, `pg_foot` and `foot`) and finally write it at the appropriate place.
 """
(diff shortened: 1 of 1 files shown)


tlienart:
Hello Mark!

Thanks for suggesting this and apologies for the delay, I was offline last week.

I could see this being useful potentially but am wondering whether it might not be better to allow `serve` to take that as a keyword argument and to pass the rendered page as well as the jd vars 

```julia
function my_function(pg, jd_vars)
  # whatever you want
end
# ...
serve(...; on_write=my_function)
# ...
```

which would be called at the same place as where you put it.

What do you think?

MarcMush:
Yes definitely, it is probably a better way of doing it
I will look into it

tlienart:
superseded  by #336

PR #336, opened 2020-01-03
Callback function `on_write`

MarcMush:
here is the second version of a callback function after each page is built and written

```julia
function my_function(pg, jd_vars)
  # whatever you want
end
# ...
serve(...; on_write=my_function)
# ...
```
_Originally posted by @tlienart in https://github.com/tlienart/JuDoc.jl/pull/309#issuecomment-566483523_

I had to add it at multiple places (that's why I went for the simpler solution in #309) but this way it's probably more user-friendly


tlienart:
Thanks a lot Mark, I'll have a detailed look asap

src/manager/judoc.jl
```diff
@@ -23,7 +23,7 @@ Keyword arguments:
 * `no_fail_prerender=true`: whether, in a prerendering phase, ignore errors and try to produce an output
 * `eval_all=false`:  whether to force re-evaluation of all code blocks
 * `silent=false`:    switch this on to suppress all output (including eval statements).
-* `cleanup=true`:    whether to clear environment dictionatires, see [`cleanup`](@ref).
+* `cleanup=true`:    whether to clear environment dictionaries, see [`cleanup`](@ref).
 """
```
tlienart: Could you please add a short  entry for `on_write` here?

tlienart:
Thanks a lot for these changes! I'll incorporate them in the coming PR #349

MarcMush:
sorry for the delay
hopefully I didn't make any mistake while changing to franklin

tlienart:
this  is now #350, thanks a  lot!

PR #349, opened 2020-01-26
Consider each page like a module (notebook style)

### LINKED ISSUE #334: There was an error running the code [Plotly]
Here's a weird one from PlotlyJS. Source is the PlotlyJS [example building script](https://github.com/sglyon/PlotlyJS.jl/blob/bc4ffc16d3d03b8644e9a9dd1f2084536a9b0807/docs/build_example_docs.jl#L62)
The following works in the REPL:
```julia
julia> using PlotlyJS

julia> a = attr(t=60, b=60, l=50, r=5)
PlotlyAttribute with fields b, l, r, and t

julia>
```
However, if I want to use this in a JuDoc page like below, I get the following error:
`There was an error running the code: ErrorException("function attr does not accept keyword arguments")`.

~~~
@def title = "Issue334"
@def hasplotly = true

## Issue

This works in the REPL, and even if I copy from `assets/pages/menu9/issue334/t1.jl`.

```julia:./issue334/t1
using PlotlyJS
a = attr(t=60, b=60, l=50, r=5)
```

\textoutput{./issue334/t1}
~~~

See online [here](https://csertegt3.github.io/JuDocPlottest/pub/menu9.html) and source [here](https://github.com/cserteGT3/JuDocPlottest/blob/master/src/pages/menu9.md).

It happens with julia 1.3.1 on windows and latest tagged versions of JuDoc and PlotlyJS:
```
(JuDocPlottest) pkg> st
    Status `C:\Users\cstamas\Documents\GIT\JuDocPlottest\Project.toml`
  [4ca9428c] JuDoc v0.4.2
  [98b081ad] Literate v2.2.1
  [47be7bcc] ORCA v0.3.1
  [f0f68f2c] PlotlyJS v0.13.0
  [d330b81b] PyPlot v2.8.2
  [44d3d7a6] Weave v0.9.1
  [9a3f8284] Random
```

### LINKED ISSUE #346: Use a single page-var dictionary
The way page vars are handled could be improved; at  the moment there's effectively:

- functions for default  global page vars  dict  and local  page vars dict
- merging that into a "current-page" page var dict which gets passed around to  functions

It would make more sense to  just have a  single globally  accessible dict that gets initialised properly and can be accessed throughout the building of the page. That would also mean fewer headaches. 

**Note**: this makes me think that it should be documented that `@def ...` are executed all in one shot and that, generally, it's safer to  have all `@def` and  `\newcommand` at the top of the page to  avoid confusing error (similar spirit than in LaTeX); adding that in the docs as a recommendation would be useful.

### LINKED ISSUE #347: use one module per page
This should eliminate issues like #334 . 

Basically every page with code  on it should be considered as one standalone notebook and be treated as such; a  very similar approach is used by Weave.  

This is ongoing work on branch `codeeval` which should be merged within a week or so.

### LINKED ISSUE #348: Live-generated plotly plots don't get all the bells and whistles
Currently the issue can be seen on the Franklin page with plotly, where the "Off-line" generated plot can zoom in/out, hover over values, etc., whereas the "On-line" generated example doesn't do anything. One difference I noted is that the "Off-line" example appears in a code block, but this is above my depth to debug in more detail. I was able to reproduce this in my own examples, however. If there is a way to "escape" the auto-generated output code block, then it would be helpful to demonstrate that as well.

https://tlienart.github.io/franklindocs/pub/extras/plotly.html

tlienart:
This is a major refactoring of the code which should hopefully significantly reduce problems when  using Franklin with code. 

* closes #346 
* closes #347 
* closes #334 and closes #348

### New stuff

* the function `fdplotly` allows you to  directly include  plotly plots (via their json representation), see https://tlienart.github.io/franklindocs/pub/extras/plotly.html#live-generated_plot for details

### Breaking changes

The PR breaks a few things but it's for the best:

* `\input{output}{...}` and `input{output:plain}{...}` are not supported anymore, prefer `\output` (just stdout in plaintext block),  `\show` (stdout+result in  plaintext block) or `\textoutput` (reprocessed stdout)
* `freezecode` is not  supported anymore (it shouldn't be necessary), unlikely  you were using this anyway
* `\file` is not supported anymore, use markdown links
* since code is ran in **modules**, paths are a bit trickier (not relative to  the script path), anyway long story short use `@OUTPUT` so for instance:

`````markdown
```julia:ex
using PyPlot
plot([1,2,3],[1,2,5])
savefig(joinpath(@OUTPUT, "ex1.png"))
```
\input{plot:1}{ex}
`````

### Reproducing stuff

For issues #334 and #348, the following:

`````markdown
@def hascode = true
@def hasplotly = true

## Plotly issue 334

```julia:plotly1
using PlotlyJS

x = y = [-2*pi + 4*pi*i/100 for i in 1:100]
z = [sin(x[i]) * cos(y[j]) * sin(x[i]*x[i]+y[j]*y[j])/log(x[i]*x[i]+y[j]*y[j]+1)
     for i in 1:100 for j in 1:100]
z_ = [z[i:i+99] for i in 1:100:10000]

data = contour(;z=z_, x=x, y=y)
plt = plot(data)

relayout!(plt, margin=attr(t=60, b=60, l=50, r=50))

html_block = html_body(plt.plot);

println("~~~")
println(html_block)
println("~~~")
```

\textoutput{plotly1}
`````

Gives me a fully working (thanks @cserteGT3 for the  example)

Screen Shot 2020-01-26 at 18 41 15

it's pannable and zoomable (@platawiec). 

It's just  not  located properly on the page but that's just because of the way `PlotlyJS.html_code` is written (it selects body and  then  tries  to just put this  big div  wherever). It would be nice to have our own function for this; I'll think about it. 

Will do  more testing then eventually merge this  as a minor release


cserteGT3:
Cool stuff! I'll give it a shot as soon as I finished my master's. (Scheduled to next tuesday 😄).
I was always confused about pathes and I like the `@OUTPUT` macro idea. Anyway I'm planning to contribute more to the ecosystem (and Franklin as well) and happy to see that Franklin attracts more and more people.

PR #415, opened 2020-03-19
Make locvar/globvar return nothing instead of erroring

fredrikekre:
Make locvar/globvar return nothing instead of erroring if the variable is not defined. Should be non-breaking since this used to error.

### DIFF
--- a/test/converter/md_defs.jl
+++ b/test/converter/md_defs.jl
@@ -87,4 +87,6 @@ end
         """ |> fd2html_td
     @test isapproxstr(F.locvar("s"), "foo bar baz etc")
     @test isapproxstr(F.locvar("s2"), "nothing")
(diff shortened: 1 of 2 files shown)


test/converter/md_defs.jl
```diff
@@ -87,4 +87,6 @@ end
         """ |> fd2html_td
     @test isapproxstr(F.locvar("s"), "foo bar baz etc")
     @test isapproxstr(F.locvar("s2"), "nothing")
+    @test F.locvar("foobar") === nothing
+    @test F.globvar("foobar") === nothing
```
fredrikekre: Bit of a weird location for these tests but atleast `locvar` was used in this testset before :man_shrugging:

tlienart:
Cool thanks!

PR #421, opened 2020-03-21
Make `fdplotly` more informative

keorn:
As discussed in Slack, makes less likely to run into issues with passing the plot object directly.


tlienart:
Thanks a lot!

PR #422, opened 2020-03-22
Allow templating blocks from Markdown

### LINKED ISSUE #355: Allow {{ var }} as short  for {{ fill  var }}
But with  least priority in case of ambiguity with a function name that would take no parameters (not many of those though: `toc`, and the `if`....`end`)

tlienart:
This allows the following

`somefile.md`:
```md
@def var = "hello"
{{var}}
```

Leading to

```html
<p>hello</p>
```

I.e. two things:

1. templating blocks can directly be used from Markdown as well as HTML
2. there is now the shortcut `{{var}}` for `{{fill var}}`, note that if `var` is the name of an internal `hfun` like `fill` or `toc`, these functions will take precedence. (closes #355) 

### Secondary modifications

- Some additional work to make Regexes easier to read + test them more. 
- Dropped the  `HBLOCK_TOC_PAT` in favour of just `HBLOCK_FUN_PAT` with zero parameters

### DIFF
--- a/src/converter/html/functions.jl
+++ b/src/converter/html/functions.jl
@@ -8,6 +8,16 @@ function convert_html_fblock(β::HFun)::String
     # try to find a function `hfun_...`
     fun = Symbol("hfun_" * lowercase(β.fname))
     isdefined(Franklin, fun) && return eval(:($fun($β.params)))
+
+    # if zero parameters, see if can fill
+    if isempty(β.params) && !isnothing(locvar(β.fname))
+        return hfun_fill([β.fname])
+    end
+
+    # XXX Future
+    # XXX see if there's an externally defined hfun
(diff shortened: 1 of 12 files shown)


fredrikekre:
`{{var}}` is a pretty nice syntax for a no-arg function `var` though

tlienart:
@fredrikekre yes, the precedence will be as follows:

1. does it exist as internal  function (i.e. part of Franklin module, you can hack this as you did)
2. is there a local variable that corresponds to it
3. is  there an externally defined `hfun` that correspond to it (this is coming feature)

So  in your case,  nothing should change  as I believe you define `hfun_*`  by evaluating  in the module.

In the future I intend to allow a `utils.jl` file where, long story  short, people can do  useful stuff among which define `hfun_*`; these would not be part  of `Franklin.` but part  of  a module `Utils.`, that's  the number (3) above.

Issue #14, opened 2018-06-18
Managing dates


tlienart:
* [ ] dates management, `Dates.today()` is no good, every time the file is recompiled (may be often). Or should just not re-interpret if there's no diff (then ok). The latter would require keeping track of the files + time of last modification in a hidden file that is gitignored.
* [ ] if empty, nothing should appear.
* [ ] dates should all have the same format (atm, it could be different).

```julia
Dates.format(now(), "U dd, yyyy")
Dates.format(Date(2018, 5, 15), "dd U yyyy")
```

See also [formats](https://en.wikibooks.org/wiki/Introducing_Julia/Working_with_dates_and_times).

PR #433, opened 2020-03-28
Add language customization to dates

lucasvreis:
This is a workaround I made for my site so I can define custom names for the months of the year and days of the week, and their abbreviations. As I am using the `locale` functionality of `Dates`, I'll be calling it "locale" customization.

The usage is simple, and like `date_format`, the user can set some local/global variables:
 - `date_days` : list with the names for days of the week
 - `date_shortdays` : list with the abbreviated names for the days of the week
 - `date_months` : list with the names of months
 - `date_shortmonths` : list with the abbreviated names of months

If any of the `short` variations are not set, the `fd_date` function tries to automatically create those by using the first three characters of the names of months or days of week.

Maybe this is not the right way to implement it, but as I have done it anyway, I present this as my first contribution :upside_down_face: I also don't know the proper way to add tests to this, should it be like the `cond-insert` tests?

### DIFF
--- a/src/utils/vars.jl
+++ b/src/utils/vars.jl
@@ -184,6 +184,8 @@ Convenience functions related to the page vars
 ============================================= =#
 
 """
+$(SIGNATURES)
+
 Convenience function taking a `DateTime` object and returning the corresponding
 formatted string with the format contained in `LOCAL_VARS["date_format"]` and 
 with the locale data provided in `date_months`, `date_shortmonths`, `date_days`, 
@@ -193,11 +195,11 @@ automatically construct them using the first three letters of the names in
 """
 function fd_date(d::DateTime)
     # aliases for locale data and format from local variables
-    format      = LOCAL_VARS["date_format"].first
-    months      = LOCAL_VARS["date_months"].first
-    shortmonths = LOCAL_VARS["date_shortmonths"].first
-    days        = LOCAL_VARS["date_days"].first
-    shortdays   = LOCAL_VARS["date_shortdays"].first
+    format      = locvar("date_format")
+    months      = locvar("date_months")
+    shortmonths = locvar("date_shortmonths")
+    days        = locvar("date_days")
+    shortdays   = locvar("date_shortdays")
     # if vectors are empty, user has not defined custom locale,
     # defaults to english
     if all(isempty.((months, shortmonths, days, shortdays)))
@@ -205,11 +207,11 @@ function fd_date(d::DateTime)
     end
     # if shortdays or shortmonths are undefined,
     # automatically construct them from other lists
-    if !isempty(days) || isempty(shortdays)
-        shortdays = [length(d) >= 3 ? d[1:3] : d for d ∈ days]
+    if !isempty(days) && isempty(shortdays)
+        shortdays = first.(days, 3)
     end
-    if !isempty(months) || isempty(shortmonths)
-        shortmonths = [length(m) >= 3 ? m[1:3] : m for m ∈ months]
+    if !isempty(months) && isempty(shortmonths)
+        shortmonths = first.(months, 3)
     end
     # set locale for this page
     Dates.LOCALES["date_locale"] = Dates.DateLocale(months, shortmonths, 
--- a/src/utils/vars.jl
+++ b/src/utils/vars.jl
@@ -14,6 +14,10 @@ const GLOBAL_VARS_DEFAULT = [
     # General
     "author"           => Pair("THE AUTHOR",   (String, Nothing)),
     "date_format"      => Pair("U dd, yyyy",   (String,)),
+    "date_months"      => Pair(String[],       (Vector{String},)),
+    "date_shortmonths" => Pair(String[],       (Vector{String},)),
+    "date_days"        => Pair(String[],       (Vector{String},)),
+    "date_shortdays"   => Pair(String[],       (Vector{String},)),
     "prepath"          => Pair("",             (String,)),
     # will be added to IGNORE_FILES
     "ignore"           => Pair(String[],       (Vector{String},)),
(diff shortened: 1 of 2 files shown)


tlienart:
Oh great, thanks a lot! I’ll review this later today.

Re tests: generally I think regression tests would be good I.e: just write a test showing what your code is supposed to do. 

The place doesn’t matter so much, but maybe `tests/utils`

Thanks again!

src/utils/vars.jl
```diff
@@ -181,9 +185,37 @@ Convenience functions related to the page vars
...
-fd_date(d::DateTime) = Dates.format(d, LOCAL_VARS["date_format"].first)
+formatted string with the format contained in `LOCAL_VARS["date_format"]` and 
+with the locale data provided in `date_months`, `date_shortmonths`, `date_days`, 
+and `date_shortdays` local variables. If `short` variations are not provided,
+automatically construct them using the first three letters of the names in
+`date_months` and `date_days`.
+"""
+function fd_date(d::DateTime)
+    # aliases for locale data and format from local variables
+    format      = LOCAL_VARS["date_format"].first
```
tlienart: the construction `LOCAL_VARS["name"].first` has a shortcut: `locvar("name")` which you could use throughout

src/utils/vars.jl
```diff
@@ -181,9 +185,37 @@ Convenience functions related to the page vars
...
+    shortdays   = LOCAL_VARS["date_shortdays"].first
+    # if vectors are empty, user has not defined custom locale,
+    # defaults to english
+    if all(isempty.((months, shortmonths, days, shortdays)))
+        return Dates.format(d, format, locale="english")
+    end
+    # if shortdays or shortmonths are undefined,
+    # automatically construct them from other lists
+    if !isempty(days) || isempty(shortdays)
+        shortdays = [length(d) >= 3 ? d[1:3] : d for d ∈ days]
```
tlienart: maybe a bit better to avoid the `?` (even though this is clearly not a bottleneck..)

```
shortdays = [d[1:min(3, length(d))] for d in days]
```

tlienart:
Your PR actually makes me realize there are a few shortcomings in Franklin in terms of how dates are processed and actually there were thoughts on this in #14 in particular, at the moment if someone does `@def date = ...` and then `{{fill date}}` it actually doesn't use the formatting. Some thoughts should go into this.

Anyway your PR is fine and it will apply to `fd_ctime` and `fd_mtime` which can be used in footer etc, that's already very useful. 

Here's a way to test your PR, kindly extend it to tests the other lists, possibly use other languages too :) 

```
s = raw"""
   @def date_format = "e, d u Y"
   @def date_months = ["janvier", "février", "mars", "avril", "mai", "juin",
                       "juillet", "août", "septembre", "octobre", "novembre", 
                       "décembre"];
   @def date_shortmonths = ["janv","févr","mars","avril","mai","juin",
                            "juil","août","sept","oct","nov","déc"];
   @def date_days = ["lundi","mardi","mercredi","jeudi",
                     "vendredi","samedi","dimanche"];
   @def reeval = true
   ```julia:ex
   #hideall
   using Franklin, Dates
   println(Franklin.fd_date(DateTime("1996-01-01T12:30:00")))
   ```
   \textoutput{ex}
   """ |> fd2html_td
@test isapproxstr(s, "lun, 1 jan 1996")
```

src/utils/vars.jl
tlienart: actually scrap that, you **must** use `first(d, 3)` otherwise you'll get in trouble if there are unicode characters

src/utils/vars.jl
```diff
@@ -181,9 +185,37 @@ Convenience functions related to the page vars
 
 """
```
tlienart: Please add `$SIGNATURES` here (I know it's my fault)

src/utils/vars.jl
```diff
@@ -181,9 +185,37 @@ Convenience functions related to the page vars
...
+    days        = LOCAL_VARS["date_days"].first
+    shortdays   = LOCAL_VARS["date_shortdays"].first
+    # if vectors are empty, user has not defined custom locale,
+    # defaults to english
+    if all(isempty.((months, shortmonths, days, shortdays)))
+        return Dates.format(d, format, locale="english")
+    end
+    # if shortdays or shortmonths are undefined,
+    # automatically construct them from other lists
+    if !isempty(days) || isempty(shortdays)
```
tlienart: This short circuit is incorrect I think. It should be

```
if !isempty(days) && isempty(shortdays)
```

(same below)

src/utils/vars.jl
```diff
@@ -181,9 +185,37 @@ Convenience functions related to the page vars
...
+        shortdays = [length(d) >= 3 ? d[1:3] : d for d ∈ days]
+    end
+    if !isempty(months) || isempty(shortmonths)
+        shortmonths = [length(m) >= 3 ? m[1:3] : m for m ∈ months]
```
tlienart: same here (the if condition + the first 3)

tlienart:
Sorry for the comments, I tested your branch locally and a few more things occurred to me along the way.

Thanks a lot for your contribution!

src/utils/vars.jl
lucasvreis: Oh, my bad, it was a short circuit like `isempty(days) || (isempty(shortdays) && do something)` before I made it into a `if` and messed things up

lucasvreis:
Where should I put the tests inside `tests/utils`, in a new file or existing one? And thank you for your comments, I'm still leaning Julia.

lucasvreis:
Soo, I added some tests and also tested it in my blog. I am able to set a language in the `config.md` file and then overwrite it with local variables in the page. But for some weird reason the japanese language test fails if I don't set a date different from the previous test, else it just reuses the output from the last test. You can see the other japanese test where I change the date runs properly. Very strange.

tlienart:
> Where should I put the tests inside `tests/utils`, in a new file or existing one? And thank you for your comments, I'm still learning Julia.

Please put them in `tests/utils/paths_vars.jl` at the end :)

test/utils/misc.jl
```diff
@@ -125,3 +125,66 @@ end
...
+    """ |> fd2html_td
+    @test isapproxstr(s, "1 janv 1996") # only short names defined
+
+    s = raw"""
+    @def date_format = "Y年ud日　E"
+    @def date_months = ["1月", "2月", "3月", "4月", "5月", "6月",
+                        "7月", "8月", "9月", "10月", "11月", "12月"];
+
+    @def date_days = ["月曜日","火曜日","水曜日","木曜日",
+                        "金曜日","土曜日","日曜日"];
```
tlienart: Cool 👍

tlienart:
Ok this is great, thanks a lot! Ignore my last comment, I'll do it, thanks again 😄

PR #434, opened 2020-03-28
I432

### LINKED ISSUE #432: Footnote becomes H2
Happens if you have one and only one empty line after a footnote before a `---` divider:
```
julia> println(Franklin.fd2html("""
       hello[^a]

       [^a]: world

       ---
       """))
<p>hello<sup id="fnref:a"><a href="#fndef:a" class="fnref">[1]</a></sup>
<h2><table class="fndef" id="fndef:a">
    <tr>
        <td class="fndef-backref"><a href="#fnref:a">[1]</a></td>
        <td class="fndef-content">world</td>
    </tr>
</table>
</h2>
```

tlienart:
closes #432 

It was a bit trickier than I thought, anyway I took the lazy route, so now if Franklin sees `---` or `***` or `___` (or more than 3) at the beginning of a new line after an empty line (so for instance `\n\n---` it will  capture it and force introduce  `<hr />`, the rest of the time it defers to Julia's markdown parser so that

```
A
---
```

is effectively  parsed as `<h2>A</h2>`.

The commit is larger than it seems but that's because I added some more tests for regexes, those can safely be ignored.

### DIFF
--- a/src/converter/markdown/blocks.jl
+++ b/src/converter/markdown/blocks.jl
@@ -23,6 +23,7 @@ function convert_block(β::AbstractBlock, lxdefs::Vector{LxDef})::AS
     βn == :LINK_DEF        && return ""
 
     βn == :DOUBLE_BRACE    && return β.ss # let HTML converter deal with it
+    βn == :HORIZONTAL_RULE && return "<hr />"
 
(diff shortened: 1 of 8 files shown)


tlienart:
Nah that's not good enough

Issue #456, opened 2020-04-17
Taxonomy


tlienart:
The kind of stuff described in https://attila.zutrinken.com/organising-content/

### Required

* a way to indicate tags, page var is fine
* some functionality to collect all tags (basically just need to go through ALL_PAGE_VARS)
* generate `mainurl/tags/tagname/` would use a `_layout/tag.html` and fill it with a generated list

This is all pretty easy to do

### Notes

* need to keep track of tags associated to all pages so that if they get modified can re-trigger generation of tag pages, could just do that in a dict.

PS: cc @wikunia since you mentioned this, if you want to give it a shot I can lend you a hand :-)

Wikunia:
Sounds nice. I'm interested on working on it on Monday. Will ask questions in slack if I have some. (I'll :smile: )

fredrikekre:
I had something like this working, but I don't remember on which machine so I guess it is lost now. Would be nice to have!

Wikunia:
One small question: this should all be done whenever such a page variable tags exists, right? So the user still just calls `serve` and all folders are created automatically.

tlienart:
Ok here's at a high level what I would do:

* optional path `_layout/tag.html` which would contain the structure that goes around one of those tag-list pages. If not there and tags are given, something would fail. 
* a local page var `tags` which allows a `Vector{String}` and is empty by default
* a global dictionary `TAGS` which contains pairs `rpath => [tags]` 
* when `process_mddefs` is called (inside `convert_md` so that's called on every page trigger) check if `tags` is given, if it is check the `TAGS` dictionary and either add, remove or update the entry. If it's identical than before, do nothing, if it isn't and that we're not on the initial fullpass then trigger a `generate_tag_pages` function

The `generate_tag_pages` function would

* take an optional argument indicating which page has triggered the update
* create a `__site/tag/` folder if it isn't there already
* invert the dictionary `TAGS` to get `tag => [rpaths]` 
* go through each tag, (if the triggering page is given & the list of `rpaths` does not intersect with it, ignore)
* for the tags that need an update, create the html which would be built out of `_layout/tag.html` and with the list of links, extracting page titles too via `{{fill title rp}}` etc
* convert that html to resolve any `{{...}}` using `convert_html` 
* create with the converted html a page `__site/tag/tagname/index.html`

The `generate_tag_pages` would also have to be called after the initial `fd_fullpass` after `ALL_PAGE_VARS` and `TAGS` is populated

[...]

PR #458, opened 2020-04-18
WIP: tags

Wikunia:
Implementation of #456 
There sometimes is a bug that the initial pass fails because `path(:site)` fails not sure why.
State:
- `TAGS` holds information `rpath => [tag1, tag2]` which gets updated in each page:
`update_tags`
- Changes get passed to `generate_tag_pages` with the optional argument `tags` which holds all changed tags. In `generate_tag_pages` the tag pages are created/updated or if the tag doesn't exist anymore removed. 
- If `generate_tag_pages` is called without the argument it does this for all tags.

The functions are probably at the wrong places :wink:.
Currently the layout of the tags page is hard coded not sure how to use a layout file for this. 

Test cases:
Need some hints how to write test cases for actual checking whether the code works. Do you actually create a franklin site in the tests to check your functions? Or just some unit tests?

### DIFF
--- a/src/converter/markdown/mddefs.jl
+++ b/src/converter/markdown/mddefs.jl
@@ -29,6 +29,7 @@ function process_mddefs(blocks::Vector{OCBlock}, isconfig::Bool)::Nothing
         push!(assignments, (String(vname) => String(vdef)))
     end
     # if in config file, update `GLOBAL_VARS` and `GLOBAL_LXDEFS`
+    rpath = splitext(locvar("fd_rpath"))[1]
     if isconfig
         set_vars!(GLOBAL_VARS, assignments)
     else
@@ -53,7 +54,78 @@ function process_mddefs(blocks::Vector{OCBlock}, isconfig::Bool)::Nothing
 
         # copy the page vars to ALL_PAGE_VARS so that they can be accessed
         # by other pages via `pagevar`.
-        ALL_PAGE_VARS[splitext(locvar("fd_rpath"))[1]] = deepcopy(LOCAL_VARS)
+        ALL_PAGE_VARS[rpath] = deepcopy(LOCAL_VARS)
     end
+    update_tags(rpath, LOCAL_VARS["tags"][1])
     return nothing
 end
+
+"""
+$(SIGNATURES)
+
+SHOULD BE PROBABLY IN A DIFFERENT FILE
+Check if tags differ from TAGS[rpath]. If yes update and call `generate_tag_pages`.
+
+**Arguments**
+
+* `rpath`:     The relative path (can also be fetched with `splitext(locvar("fd_rpath"))[1]`)
+* `tags`:      List of tags
+"""
+function update_tags(rpath::String, tags::Vector{String})
+    # if the path had tags before
+    if haskey(TAGS, rpath)
+        if TAGS[rpath] != tags
+            union_tags = union(TAGS[rpath], tags)
+            TAGS[rpath] = tags
+            generate_tag_pages(;tags=union_tags)
+        end
+    else
+        TAGS[rpath] = tags
+        generate_tag_pages(;tags=tags)
+    end
+end
+
+tag_line(url, title) = "<li><a href='$url'>$title</li>"
+
+"""
+$(SIGNATURES)
+
+SHOULD BE PROBABLY IN A DIFFERENT FILE
+Create and/or update `__site/tag` folders/files
+
+**Optional Arguments**
+
+* `tags`:      List of changed tags. If `nothing` all tags will be updated
+"""
+function generate_tag_pages(;tags=nothing)
+    isempty(TAGS) && return
+    tags == String[] && return 
+    !isdir(joinpath(path(:site), "tag")) && mkdir(joinpath(path(:site), "tag"))
+    # tag => [rpath]
+    inv_tags_dict = invert_dict(TAGS)
+    tags === nothing && (tags = keys(inv_tags_dict))
+    
+    layout_key  = ifelse(FD_ENV[:STRUCTURE] < v"0.2", :src_html, :layout)
+    layout      = path(layout_key)
+    head        = read(joinpath(layout, "head.html"),      String)
+    pg_foot     = read(joinpath(layout, "page_foot.html"), String)
+    foot        = read(joinpath(layout, "foot.html"),      String)
+    
+    for tag in tags
+        dir = joinpath(path(:site), "tag", tag)
+        !isdir(dir) && mkdir(dir)
+        fpath = joinpath(dir, "index.html")
+        if !haskey(inv_tags_dict, tag)
+            rm(fpath)
+            rm(dir)
+        else
+            tag_content = "<h1>Tag: $tag</h1>\n"
+            for rpath in inv_tags_dict[tag] 
+                tag_content *= tag_line(rpath, pagevar(rpath, "title"))*"\n"
+            end
+            content = build_page(head, tag_content, pg_foot, foot)
+            write(fpath, convert_html(content))
+        end
+    end
+end
+
--- a/src/utils/vars.jl
+++ b/src/utils/vars.jl
@@ -8,6 +8,10 @@ for that value. (e.g.: "THE AUTHOR" => (String, Nothing))
 """
 const GLOBAL_VARS = PageVars()
 
+# rpath => tags
+# UPPERCASE EVEN IF NOT CONST?
+TAGS = Dict{String, Vector{String}}()
+
 const GLOBAL_VARS_DEFAULT = [
     # Folder org
     "folder_structure" => Pair(FD_ENV[:STRUCTURE], (VersionNumber,)),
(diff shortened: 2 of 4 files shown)


src/converter/markdown/mddefs.jl
```diff
@@ -53,7 +54,78 @@ function process_mddefs(blocks::Vector{OCBlock}, isconfig::Bool)::Nothing
 
         # copy the page vars to ALL_PAGE_VARS so that they can be accessed
         # by other pages via `pagevar`.
-        ALL_PAGE_VARS[splitext(locvar("fd_rpath"))[1]] = deepcopy(LOCAL_VARS)
+        ALL_PAGE_VARS[rpath] = deepcopy(LOCAL_VARS)
     end
+    update_tags(rpath, LOCAL_VARS["tags"][1])
```
tlienart, line 59: use `locvar("tags")`

src/converter/markdown/mddefs.jl
```diff
@@ -53,7 +54,78 @@ function process_mddefs(blocks::Vector{OCBlock}, isconfig::Bool)::Nothing
...
+* `rpath`:     The relative path (can also be fetched with `splitext(locvar("fd_rpath"))[1]`)
+* `tags`:      List of tags
+"""
+function update_tags(rpath::String, tags::Vector{String})
+    # if the path had tags before
+    if haskey(TAGS, rpath)
+        if TAGS[rpath] != tags
+            union_tags = union(TAGS[rpath], tags)
+            TAGS[rpath] = tags
+            generate_tag_pages(;tags=union_tags)
```
tlienart, line 80: this is a bit weird, why would you do the `union`? just take `tags` and discard the previous ones right?

src/converter/markdown/mddefs.jl
```diff
@@ -53,7 +54,78 @@ function process_mddefs(blocks::Vector{OCBlock}, isconfig::Bool)::Nothing
...
+        end
+    else
+        TAGS[rpath] = tags
```
tlienart, line 83: this `else` branch is not necessary, just return nothing. If the tags haven't changed then the tag pages don't need to be updated.

src/converter/markdown/mddefs.jl
```diff
@@ -53,7 +54,78 @@ function process_mddefs(blocks::Vector{OCBlock}, isconfig::Bool)::Nothing
...
+        generate_tag_pages(;tags=tags)
+    end
+end
+
+tag_line(url, title) = "<li><a href='$url'>$title</li>"
```
tlienart, line 88: for composability: 
```
tag_line(url, title) = "<li>" * html_ahref(url, title; class="tag-link") * "</li>"
```

src/converter/markdown/mddefs.jl
```diff
@@ -53,7 +54,78 @@ function process_mddefs(blocks::Vector{OCBlock}, isconfig::Bool)::Nothing
...
+$(SIGNATURES)
+
+SHOULD BE PROBABLY IN A DIFFERENT FILE
+Create and/or update `__site/tag` folders/files
+
+**Optional Arguments**
+
+* `tags`:      List of changed tags. If `nothing` all tags will be updated
+"""
+function generate_tag_pages(;tags=nothing)
```
tlienart, line 100: for now don't do this keyword `tags=`, one would have to be fairly careful to make sure it works as intended.

src/converter/markdown/mddefs.jl
```diff
@@ -53,7 +54,78 @@ function process_mddefs(blocks::Vector{OCBlock}, isconfig::Bool)::Nothing
...
+    isempty(TAGS) && return
+    tags == String[] && return 
+    !isdir(joinpath(path(:site), "tag")) && mkdir(joinpath(path(:site), "tag"))
```
tlienart, line 103: add a branch `if FD_ENV[:STRUCTURE] < v"0.2"` in which case it should be `joinpath(path(:pub), "tag")` (likely  what causes your bug)

src/converter/markdown/mddefs.jl
```diff
@@ -53,7 +54,78 @@ function process_mddefs(blocks::Vector{OCBlock}, isconfig::Bool)::Nothing
...
+    tags === nothing && (tags = keys(inv_tags_dict))
+    
+    layout_key  = ifelse(FD_ENV[:STRUCTURE] < v"0.2", :src_html, :layout)
+    layout      = path(layout_key)
+    head        = read(joinpath(layout, "head.html"),      String)
+    pg_foot     = read(joinpath(layout, "page_foot.html"), String)
+    foot        = read(joinpath(layout, "foot.html"),      String)
+    
+    for tag in tags
+        dir = joinpath(path(:site), "tag", tag)
```
tlienart, line 115: again fork on folder structure, probably easier to set a `tag_path` earlier that is set to  either `joinpath(path(:site), "tag")` or to `joinpath(path(:pub), "tag")`

src/converter/markdown/mddefs.jl
```diff
@@ -53,7 +54,78 @@ function process_mddefs(blocks::Vector{OCBlock}, isconfig::Bool)::Nothing
...
+        !isdir(dir) && mkdir(dir)
+        fpath = joinpath(dir, "index.html")
```
tlienart, line 117: you can use `mkpath(joinpath(dir, "index.html"))` which will always work

src/converter/markdown/mddefs.jl
```diff
@@ -53,7 +54,78 @@ function process_mddefs(blocks::Vector{OCBlock}, isconfig::Bool)::Nothing
...
+        if !haskey(inv_tags_dict, tag)
+            rm(fpath)
+            rm(dir)
```
tlienart, line 120: just the second one is needed + you need to  add `recursive=true`

src/converter/markdown/mddefs.jl
```diff
@@ -53,7 +54,78 @@ function process_mddefs(blocks::Vector{OCBlock}, isconfig::Bool)::Nothing
...
+        else
+            tag_content = "<h1>Tag: $tag</h1>\n"
```
tlienart, line 122: nitpicking: it's better practice to user a  buffer here.

`buf = IOBuffer()` then write iteratively to the buffer `write(buf, tag_line...)` finally pass it with `String(take!(buf))` to `build_page`

src/utils/vars.jl
```diff
@@ -8,6 +8,10 @@ for that value. (e.g.: "THE AUTHOR" => (String, Nothing))
 """
 const GLOBAL_VARS = PageVars()
 
+# rpath => tags
+# UPPERCASE EVEN IF NOT CONST?
+TAGS = Dict{String, Vector{String}}()
```
tlienart, line 13: just make it const (you can still modify it bc it's  a `Dict`)

src/converter/markdown/mddefs.jl
Wikunia: I need the union such that the old tags get updated (the current page should be removed from the list there)

src/converter/markdown/mddefs.jl
Wikunia: the `else` is for the page has tags now and hasn't had tags before.

src/converter/markdown/mddefs.jl
Wikunia: haven't seen that function :smile: . Will change

src/converter/markdown/mddefs.jl
Wikunia: So just update all tags?

src/converter/markdown/mddefs.jl
Wikunia: Wanted to make sure that I don't delete more than intended but depends on the definition of intended :laughing:

src/utils/vars.jl
Wikunia: oh I was confused as I didn't know that.

src/converter/markdown/mddefs.jl
tlienart: sorry I don't get this. Say you had

```
@def tags = ["maths", "fun"]
```

and that you then change it to

```
@def tags = ["maths", "notfun"]
```

you should have `TAGS[rpath] == ["maths", "notfun"]` right? not `["maths", "fun", "notfun"]` 

am  I not seeing something here?

src/converter/markdown/mddefs.jl
tlienart: ok yes

src/converter/markdown/mddefs.jl
tlienart: yes, it will take negligible time anyway, and we want to ensure that the basic behaviour works as intended first. 

Then  basically  what you want to know is what page has changed and then the sublist of tag pages that you need to update. I think it's a bit more subtle than the way you've done it here

src/converter/markdown/mddefs.jl
Wikunia: It's only for `generate_tag_pages` which needs the old ones to update the "fun" page as well.

src/converter/markdown/mddefs.jl
tlienart: since you're in `__site` everything you'd be deleting would have been generated by Franklin in the first place so little risks :)

src/converter/markdown/mddefs.jl
Wikunia: Somehow `PATHS` is empty in the first run. Not sure at which stage `set_paths!` is actually get called.

src/converter/markdown/mddefs.jl
Wikunia: This way `tags` don't get removed, right? In my case I remove the tags when there is a tag which existed and is therefore in `tags` but not in `TAGS`. In your case one would need to delete the `tag` folder and rebuild it?

src/converter/markdown/mddefs.jl
tlienart: hmm ok yes, I think I'm with you now but I think that should be done in generate pages; I'll have a play with your branch and see

Wikunia: Ok let me know if there is something I can do (if that helps :smile: )

tlienart:
this is integrated in #461 and, because I suck, I wasn't able to push the changes directly here

PR #461, opened 2020-04-22
[wip] Taxonomy

tlienart:
cc: @Wikunia this is based on your PR #458 ; I've tested it briefly and it seems ok but I need to write fuller tests for it. 

In the mean time please have  a  look and let me know  what you think! and thanks again for looking into this :) taxonomy and chronological pagination are the two big missing features remaining before we can start copying everyone's themes 

* [x] basic functionality ok
* [x] tests
  * [x] adding, removing tags
  * [x] tag page generation and deletion
  * [x] dict inversion
* [x] make tag line customisable (hfun) 
  * [x] test

**Note** @Wikunia, btw I think you were completely right to build the page with `head` and `foot`, it's of course much simpler, and someone could just do  `{{ispage /tag/*}} ... {{end}}` to toggle a specific layout for those pages :) :+1:


src/converter/markdown/mddefs.jl
```diff
@@ -53,7 +54,85 @@ function process_mddefs(blocks::Vector{OCBlock}, isconfig::Bool)::Nothing
...
+
+    # Get the dictionary tag -> [rp1, rp2...]
+    TAG_PAGES = invert_dict(PAGE_TAGS)
+    all_tags  = collect(keys(TAG_PAGES))
+
+    # check if the tag dir is there
+    isdir(path(:tag)) || mkdir(path(:tag))
+    # cleanup any page that may still be there but shouldn't
+    for dname in readdir(path(:tag))
+        dname in all_tags || rm(joinpath(path(:tag), f), recursive=true)
```
tlienart, line 115: should be `rm(joinpath(path(:tag), dname), recursive=true)`

tlienart:
Ok this is now in 0.7.6, un-announced officially because I'd like to battletest it a bit more (I'll add a theme/template which uses it soon). Help to report any problems in the mean time would be very welcome here!

### API

* add tags to a page with `@def tags = ["aa", "bb"]` 
* this will generate pages at `/tag/aa/` and `/tag/bb/` with the list of pages with the same tags
* the way the list of results is shown is controlled by the html function `hfun_list` which
  - finds the pages with a given tag
  - sorts them by date, and if date not given by creation date
  - puts the results in a simple list

Users can easily overwrite this behaviour, they would need to implement something like this in their `utils.jl`, modifying the core of the logic however they want. 

For instance users could define a variable `short_description` for all these pages and display  those in some order instead of a list.

```julia
function hfun_list(params)
    tag = params[1]
    TAG_PAGES = globvar("fd_tag_pages")
    c = IOBuffer()
    # -------------------------------------------
    # add your logic here
    write(c, "<h1>Tag: $tag</h1>")
    write(c, "<ul>")
    rpaths = TAG_PAGES[tag]
    sorter(p) = begin
        pvd = pagevar(p, "date")
        if isnothing(pvd)
            return Date(Dates.unix2datetime(stat(p * ".md").ctime))
        end
        return pvd
    end
    sort!(rpaths, by=sorter, rev=true)
    for rpath in rpaths
        title = pagevar(rpath, "title")
        if isnothing(title)
            title = "/$rpath/"
        end
        write(c, "<li><a href=\"/$rpath/\">$title</a></li>")
    end
    write(c, "</ul>")
    # -------------------------------------------
    return String(take!(c))
end
```

PR #481, opened 2020-05-08
Taxonomy (second attempt)

tlienart:
Alright folks, this brings **tags** (back) in. When I have gained more confidence that this works as expected I'll merge and it will be minor release `0.8`. This is one of the main missing feature that I really wanted to add in Franklin.

I'd like some help to check it has the right behaviour. Basic stuff is unit-tested but the overall working of the tags is a bit subtle in terms of cleaning things up when tags are added/removed while the server is running. 

### To test this

Just add `@def tags = [...]` on meaningful pages. The first run will force-add a basic `_layout/tag.html` page which will be the mould for all generate tag pages. You should tweak this so that it respects the style of the rest of your pages. In the future it will be part of templates. 
The function `{{taglist}}` could be replaced by your own function `{{customTaglist}}` by adding a `hfun_customTaglist` in your `utils.jl` but let's maybe not worry about that right now.

The basic one `{{taglist}}` just adds the list of pages corresponding to  the tag by reverse chronological order. 

The very important thing to note is that `_layout/tag.html` **must not depend on local page variable**. It can only:

* use global page variables
* use page variable via `{{fill var path/to/page}}` 

The only  exception is `fd_tag` which contains the tag name. This you can use in the title for instance. For convenience let's reproduce the default HTML with some comments:

```html
<!doctype html>
<html lang="en">
<head>
  <meta charset="UTF-8">
  <meta name="viewport" content="width=device-width, initial-scale=1">
  <title>Tag: {{fill fd_tag}}</title>
</head>
<body>
  <div class="{{div_content}} tagpage">
    {{taglist}}
  </div>
</body>
</html>
```

* You can see the `{{fill fd_tag}}` this will fill the tag name
* You can  see the `{{taglist}}` function which will generate something like

```html
<ul>
  <li><a href="/path/to/page_more_recent/">Title if it exists</a></li>
  <li><a href="/path/to/page_less_recent/">Title if it exists</a></li>
</ul>
```

Finally you can still use `{{ispage /tag/tagname/}}...{{end}}` to change the style depending on which tag page you're on.


Wikunia:
Some comments:
I think the default `tag.html` in `_layout` should include the head and foot.
My setup:
```
{{insert head.html }}
  <div class="{{div_content}} tagpage">
    {{taglist}}
  </div>
{{insert foot.html}}
```

The links doesn't work for me i.e
A `href` linking to `/blog/constraint-solver-table-constraint/index/` is created but it should be `/blog/constraint-solver-table-constraint/` or with index.html in the end.

Adding, deleting tags seem to work for me ;)

tlienart:
Thanks for the feedback! 

* link ending with `index`, yes I noticed that, simple fix coming
* adding head  and foot, so actually that cannot be the case because tag pages **must not** depend on local page var but, by default,  `head` and  `foot` will typically  have things like `{{fill title}}` etc which are ambiguous in generated pages. 

The new templates show how one can do this (e.g.: https://github.com/tlienart/FranklinTemplates.jl/tree/tags/src/templates/basic/_layout), basically make the content of `head` more modular, and plug those blocks also in `tag.html`.

Wikunia:
I see yeah I've changed my foot and head accordingly but your decision makes sense then I guess.
The fix doesn't seem to have worked for me. I did `] add Franklin#tags2` again hope that should be enough

tlienart:
Sorry the initial fix was something else, the actual fix will be in soon, thanks!

tlienart:
Now in 😄

tlienart:
I've updated all templates, will play with this a bit more and merge tomorrow I think.

Wikunia:
Haha now the href is just `/` :rofl:

tlienart:
oops

tlienart:
fixed... dumb typo

src/utils/html.jl
```diff
@@ -141,7 +141,7 @@ function get_url(rpath)
         url = rpath
     end
     if endswith(url, "index")
-        url = url[1:length(url)-length(url)]
```
Wikunia: That's the kind of bugs I do all the time but sometimes I notice them way too late.
Did `if a < length(a)` before where a was a number just to find out that `length(Number) == 1 ` obviously I meant `if a < length(b)`...

tlienart: I guess that's what  Unittests are for... I still need to add a couple more before merging :)

Wikunia: Yeah I should do some of those. Mostly writing end to end tests. They are way more fun but well they don't always catch those if your code sneaks around it and it still works when the else branch is used.

newptcai:
How to check if a page has a tag in templates?

tlienart:
@newptcai please post in issues not in PR, tags will be advertised in the NEWS section soon. `{{ispage /tag/tagname/}}` works.

PR #488, opened 2020-05-13
no method matching findfirst(::Char, ::String)

guberger:


tlienart:
are you on Julia 1.0 ? if so Franklin does not support it and so uses a bunch of things that have been brought in later such as that find first; I have

```julia-repl
julia> findfirst('{', "a{b")
2
```

guberger:
I am on Julia 1.2. Which version of Julia do you recommend? The latest one, 1.4.1?

tlienart:
Anything after 1.3 should be fine.

```
 │ Julia 1.3
  │
  │  This method requires at least Julia 1.3.

  Examples
  ≡≡≡≡≡≡≡≡≡≡

  julia> findfirst('a', "happy")
  2
  
  julia> findfirst('z', "happy") === nothing
  true
```

I'll adjust the docs to reflect this, I thought 1.1 would be fine but in fact it's >= 1.3

guberger:
It works with 1.4.1. Thanks

PR #504, opened 2020-05-29
website_ -> site_ vars

zdroid:
`site` is way shorter than `website`. It also allows longer var names like `site_description` instead of `website_descr` to be used.

Jekyll uses `site` as one of the variable scopes, so that may make the transitions easier.

In case this gets accepted, https://github.com/tlienart/FranklinTemplates.jl should be updated as well.


tlienart:
I’m favourable to this but would rather it was an alias so that it doesn’t break what people like cormullion and others (including julialang) have otherwise it’d be a breaking release. 

I’m not exactly sure how to do aliasing cleanly so that it doesn’t become a big fat mess though :-) let’s maybe wait until either we have a way to do aliasing nicely or we do a breaking release. (Or we do something like versioning for variable names, a bit similar to what I did with the file paths, and people can opt out of the “new” namings)

zdroid:
This may actually be moved to `site` variable scope (instead of having a variable prefix). That would be backwards-compatible too, since `website_` doesn't have to be removed.

tlienart:
I'll close this for now then thanks!

PR #508, opened 2020-06-03
update upper bounds of LiveServer

cserteGT3:
LiveServer compat is still on "0.3", is there any compatibility reason for it? If not, then this PR updates compat section to support 0.4 and 0.5 as well.


cserteGT3:
Ohh, I didn't see #508 and CompatHelper. Sorry about that.

Issue #506, opened 2020-06-01
Use LiveServer 0.4 with host keyword


tlienart:
https://github.com/asprionj/LiveServer.jl/pull/98

expose the `host` keyword to  `serve` for those who use this remotely

tlienart:
this is done

PR #510, opened 2020-06-03
use LiveServer with host keyword

cserteGT3:
Implements #506 
I've only done the easy part, added the keyword to the function. I've tested it with my laptop and phone and worked well.
I added the new kwarg to the docstrings, should the documentation mention it somewhere?

Drops the support for LiveServer 0.3, and adds for 0.4,0.5. With the compat section updated, makes #507, #508, #509 unnecessary.


tlienart:
Hahaha thanks a lot for that! sorry for guilt-tricking you into coming back here.

This looks like it's exactly what's required, I'll be a bit slower in merging this as I want to be 100% sure of it but thanks  a lot already!

PR #511, opened 2020-06-04
WIP: basic way to ignore files with *

Wikunia:
Just a first approach. I actually decided to have a joker `*` which can be at any position such that `posts/*/post.md` works for example.

Problems with current approach:
- Generating the regex is not super nice maybe there is a way to escape chars. 
- The regex should probably be generated once in the `scan_input_dir!` function instead which would change `Vector{String}` to `Vector{Union{String, Regex}}`.

### DIFF
--- a/src/manager/dir_utils.jl
+++ b/src/manager/dir_utils.jl
@@ -278,8 +278,19 @@ function should_ignore(fpath::AS, files2ignore::Vector{String},
     else
         fpath = fpath[length(path(:folder))+length(PATH_SEP)+1:end]
     end
-    flag  = findfirst(c -> c == fpath, files2ignore)
-    isnothing(flag) || return true
+    files2ignore_regex = []
+    for ignore in files2ignore
+        if occursin("*", ignore)
+            ignore_str = replace(ignore, "." => "\\.")
+            ignore_str = replace(ignore_str, "*" => ".*?")
+            ignore_regex = Regex("^"*ignore_str*"\$")
+            push!(files2ignore_regex, ignore_regex)
+        else
+            push!(files2ignore_regex, ignore)
+        end
+    end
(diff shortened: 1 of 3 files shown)


fredrikekre:
How about letting users put regex in `ignore` directly instead? E.g.
```
@def ignore = ["some string", r"some regex"]
```

tlienart:
I was a bit worried that this might be more bug prone but looking at the code it seems like it would be a natural progression anyway so might as well, what do you think @Wikunia ?

Wikunia:
Sounds good to me. People who want that behavior are probably the ones who know enough about regex to make it work + having one in the docs just to show seems reasonable.
Will change the PR.

Wikunia:
One problem:
I don't know what's the best way to allow regex for directories or whether that is needed at all.

tlienart:
Thanks a lot Ole, I'll try to have a look at this later today!

fredrikekre:
>  I don't know what's the best way to allow regex for directories or whether that is needed at all.

Just a regex that matches the path? For example
```
r"posts/.*"
```
would ignore everything in that folder.

Wikunia:
Yeah currently there is splitting between files and folders and for folders it only checks whether it starts with the folder and for files there is an exact match. 
Probably one can just do the same as for files and just check if it is a regex or not and if it's a string test `startwith` and otherwise match regex.

tlienart:
> > I don't know what's the best way to allow regex for directories or whether that is needed at all.
> 
> Just a regex that matches the path? For example
> 
> ```
> r"posts/.*"
> ```
> 
> would ignore everything in that folder.

note that you can already ignore an entire folder by just adding a `/` so 

```
@def ignore = ["path/to/dir/", "path/to/file.md"]
```

will ignore the entire dir and the specific file.

Issue #512, opened 2020-06-05
Remove \underline command to avoid clash with KaTeX


tlienart:
Either

* remove it (prefer \style) 
* keep it and disable it in math mode

Issue #494, opened 2020-05-17
Allow exclusion of files from minification step


tlienart:
It was brought to my attention (@tbeason, thanks!) that the minification step  of already-minified assets can cause problems (e.g. `bootstrap.min.css`). It would seem reasonable to just not touch these assets as they, more likely than not, have been expertly optimized. 

A kwarg `skip_minify = [...]` could do the trick in optimize, a global var `@def =  skip_minify = []`  (like ignore)

zdroid:
What problems exactly are caused? A slight modification of the minify script can solve this easily (i.e. by ignoring the `.min.css` / `.min.html` files).

tbeason:
In my case, the full CSS stylesheet was not being applied. Some magically stuff worked, other stuff didn't. I also think the easiest way to fix the problem would be to just always ignore files that are already minimized as indicated by their names + a short mention in the docs about it.

zdroid:
@tbeason Could you provide more details? It'd be good to know what worked and what didn't so that all possible errors are avoided. :smile:

tbeason:
One thing that for sure didn't work was that the Bootstrap table styles I was trying to apply would not work properly (like row hovering or table striping).

I don't think it really matters much though because the right thing to do is to just not try to minify a file that has already been minified.

tlienart:
@zdroid the minification script I'm using here is a python script that's released under LGPL3 so I'm a bit wary of modifying it myself (assuming that's what you suggest), though given I _call_ it (like a bash call), I could modify that call to avoid specific file patterns which would not touch the code. 

I haven't tried to reproduce the problem but I would still think that trying to minify an already-minified file is not very clever, whether it can lead to bugs is maybe surprising but it still seems like a good idea to not do it what do  you think?

zdroid:
Right. Perhaps we can leave them unminified by default, but also add an option to minify them.

PR #537, opened 2020-06-23
Misc

tlienart:
closes a few simple issues, updates NEWS to indicate what has changed that is noteworthy. Essentially:

* dropping `\underline` (use `\style` instead) #512 
* already minified files (`*.min.css`) are left untouched #494 
* adding an `{{isempty xxx}}` and `{{isnotempty xxx}}` conditional block #514


tlienart:
well that was embarrassing, got there in the end 🙈

PR #557, opened 2020-07-16
Simple redirects

tlienart:
```
{{redirect some/path.html}}
```

will add `[site]/some/path.html` and link to the page that has the hfun.

assumption: `some/path.html` does not clash (if it doesn't, whatever exists will stay as is and the redirect will not be created).

### DIFF
--- a/src/converter/html/functions.jl
+++ b/src/converter/html/functions.jl
@@ -219,3 +219,39 @@ function hfun_taglist()::String
     write(c, "</ul>")
     return String(take!(c))
 end
+
+
+"""
+$(SIGNATURES)
+
+H-Function of the form `{{redirect /addr/blah.html}}`.
+"""
+function hfun_redirect(params::Vector{String})::String
+    if length(params) != 1
+        throw(HTMLFunctionError("I found an {{redirect ...}} block and expected a single " *
+                                "address but got $(length(params)). Verify."))
+    end
+    addr = params[1]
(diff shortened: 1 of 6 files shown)


src/converter/html/functions.jl
```diff
@@ -219,3 +219,39 @@ function hfun_taglist()::String
...
+        dst = joinpath(path(:site), addr)
+    end
+    isfile(dst) && return ""
+    mkpath(splitdir(dst)[1])
+    write(dst, """
+    <!-- Generated Redirect -->
+    <!doctype html>
+    <html>
+    <head>
+      <meta http-equiv="refresh" content="0; url=$(locvar(:fd_url))">
```
shashi, line 252: should this prefix with `prepath`? it currently does not...

tlienart: ah, crap, so the error is not here but there: https://github.com/tlienart/Franklin.jl/blob/d44d6d0d76cf9656e483146d71f52f8dd2f37448/src/converter/html/link_fixer.jl#L82 the regex should include `url` 

I'll fix this asap; thanks for the heads up

tlienart: Fixed in the (pending) release 0.9.16

shashi: Thank you!

shashi: I see how it's done. Cool!

PR #566, opened 2020-07-27
Add the ability to `publish` without doing the `git push`

DilumAluthge:
Sometimes I want to do all of the things in `publish` except for the `git push` part.

This PR allows a user to skip the `git push` while still doing everything else in `publish`.


tlienart:
So then effectively it's the same as `optimize` right? But I understand that this may be nice to only have one command to use, thanks!

DilumAluthge:
> So then effectively it's the same as `optimize` right? But I understand that this may be nice to only have one command to use, thanks!

Except you can pass `; final = lunr`, and the `prepath` will get set correctly automatically, etc.

tlienart:
ah yes right!

PR #572, opened 2020-07-31
Bugfix for the logic in is(not)def

fredrikekre:


fredrikekre:
I will push a test too.

src/converter/html/blocks.jl
```diff
@@ -64,8 +64,8 @@ function process_html_cond(hs::AS, qblocks::Vector{AbstractBlock},
         lag = 1
         if !(βi isa Union{HIsPage, HIsNotPage})
             k = haskey(LOCAL_VARS, βi.vname)
-            if !k
-                k = βi isa HIsNotDef
+            if βi isa Union{HIsDef, HIsNotDef}
+                k &= βi isa HIsDef
```
fredrikekre, line 68: Eh, this is not correct either.

tlienart:
Thanks!!

PR #577, opened 2020-08-11
Add Julia 1.5 testing

zdroid:
Maybe 1.3 should be dropped too, since it isn't supported anymore (neither LTS nor stable release). (But it's the lower bound in `Project.toml`.)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,6 +5,7 @@ os:
 julia:
   - 1.3
   - 1.4
+  - 1.5
   - nightly
 matrix:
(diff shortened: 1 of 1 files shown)


tlienart:
I think I'd like 1.3 to stay there bc it's in the docs as the lowest requirements though of course I expect people will have higher versions. Could you remove 1.4 though? that's useless now if there's 1.3 and 1.5. Thanks!

zdroid:
Yeah, good idea. Done.

PR #583, opened 2020-08-14
Removing redundant unique call before making Set

mcognetta:
Removing a call to `unique` before making a Set of the local tags.


tlienart:
Thank you!

PR #590, opened 2020-08-17
support for +++...+++ blocks, closes #587

### LINKED ISSUE #587: Support TOML-like front-matter like block for page var def
It seems some people would find that useful, in effect it would just be an additional block where

```
+++
a = 5
b = [1,2,3]
+++
Content
```

Would be re-parsed as

```
@def a = 5
@def b = [1,2,3]

Content
```

This should be easy to do, the only thing to look at is whether it ~~clashes with the definitions of either the horizontal rule~~ or ~~the title stuff~~ (**edit**, no that's with `---` or `===` so if we support `+++` it's fine); ~~maybe that can be avoided by only checking for this block at the very start of the document~~.

tlienart:
This is now allowed:

```
+++
a = 5
b = "hello"
+++
{{b}} {{a}}
```

(will show `hello 5`)

and can be used instead of

```
@def a = 5
@def b = "hello"
{{b}} {{a}}
```

**Note**: what's in the `+++...+++` block is executed as raw Julia code and the assignments are then passed to GLOBAL_VARS or LOCAL_VARS depending on whether we're on `config.md` or not. This means that you could write arbitrary code and save the output in a var which is then accessible for instance:

```
+++
foo(x, y, z) = x^2 + 2y + z
out = foo(2,3,4)
+++
{{out}}
```

will show `14`. 

(if you want to show the code you should use evaluated code blocks as per usual though)

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "Franklin"
 uuid = "713c75ef-9fc9-4b05-94a9-213340da978e"
 authors = ["Thibaut Lienart <tlienart@me.com>"]
-version = "0.9.4"
+version = "0.9.5"
 
 [deps]
 Crayons = "a8cc5b0e-0ffa-5ad4-8c14-923d3ee1735f"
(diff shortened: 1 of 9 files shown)


tlienart:
@ ordovician apologies for the ping, I thought you would find that interesting.

ordovician:
Cool! That was fast

PR #608, opened 2020-09-05
[WIP] Adding a path var for a custom tag page.

### LINKED ISSUE #604: allow custom tag address
At the moment tags will be at `/tag/XXX`;  could be good to  offer the possibility for `/YYY/XXX` via a global var (e.g. if  people  want `/tags/` instead  of  `/tag/`

mcognetta:
This resolves #604 by allowing a user to set a `tag_page_path` variable in `config.md` which changes the page /tag/XXX to /YYY/XXX where YYY is given by the user. I open it now as a WIP to check that the code part of the change is correct before adding documentation/tests/etc.

TODO:
- add tests
- add to documentation


tlienart:
Yes this is perfect thanks!

Did you need to add the resets to global var to get the tests to pass?

I'm travelling atm but will aim to merge asap, thanks again!

mcognetta:
Yes, I found I needed to set them as otherwise the tag_page_path could not be found. I think that I can set them just once per file though, rather than on a test-by-test basis like I have now. I will experiment and see.

mcognetta:
I reduced the number of calls to `def_GLOBAL_VARS!()` in the tests. It was sufficient to call it once at the start of the files in most cases.

tlienart:
Reviewing now and will then merge

PR #652, opened 2020-10-13
Remove ref to franklindocs repo

remi-garcia:
I found three links to an archived repo.

I am not sure about the change in the README.md (source -> see docs/). What do you think?

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,7 +35,7 @@ Go to [Franklin's main website](https://franklinjl.org).
 
 Some examples of websites using Franklin (_if you're using Franklin with a public repo, consider adding the "franklin" tag to the repo to help others find examples, thanks!_)
 
-* Franklin's own website is written in Franklin, [source](https://github.com/tlienart/franklindocs)
+* Franklin's own website is written in Franklin, [see docs/](docs/)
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -35,7 +35,7 @@ Go to [Franklin's main website](https://franklinjl.org).
 
 Some examples of websites using Franklin (_if you're using Franklin with a public repo, consider adding the "franklin" tag to the repo to help others find examples, thanks!_)
 
-* Franklin's own website is written in Franklin, [source](https://github.com/tlienart/franklindocs)
+* Franklin's own website is written in Franklin, [see docs/](docs/)
```
tlienart, line 38: I don't think that works, https://github.com/tlienart/Franklin.jl/tree/master/docs is probably safer

tlienart (changes requested):

tlienart:
great, thanks!

README.md
fredrikekre: I think it works, and has the additional benefit that it links to the correct commit, but perhaps you always want people to see the `master` branch?

tlienart: no that's great, I learned something!

PR #653, opened 2020-10-15
Add robots.txt

### LINKED ISSUE #651: Robots.txt and others
It would be nice to have a straightforward way to generate:
- [ ] [robots.txt](https://www.robotstxt.org/)
- [ ] [security.txt](https://securitytxt.org/)
- [ ] [humans.txt](http://humanstxt.org/)
- [ ] etc.?

<details>
  <summary>Other stuff</summary>
Thanks for the package, I'm trying it right now and it is really comfortable to use.
</details>

remi-garcia:
First step to resolve #651 

To do before merging:
- [x] Add a way to disallow robots

How it is done elsewhere:
- Jekyll generates its `robots.txt` with the `sitemap` generation: see [https://github.com/jekyll/jekyll-sitemap](https://github.com/jekyll/jekyll-sitemap)
- Hugo generates `robots.txt` has a template(?). I am not sure I understood the way it works. However I am not familiar with any of the involved technologies and it might be straightforward for others. See [https://github.com/gohugoio/hugo/blob/master/docs/content/en/templates/robots.md](https://github.com/gohugoio/hugo/blob/master/docs/content/en/templates/robots.md)

As a `sitemap_generator` is present in Franklin, I went for a `robots_generator` that include a line that links to sitemap.xml.
See [here](https://en.wikipedia.org/wiki/Robots_exclusion_standard#Sitemap) or [here](https://www.sitemaps.org/protocol.html#submit_robots) for more information on `Sitemap:` keyword.


src/manager/robots_generator.jl
```diff
@@ -0,0 +1,28 @@
...
+$SIGNATURES
+
+Generate a `robots.txt`, if one already exists, it will be replaced.
+"""
+function robots_generator()
+    dst = joinpath(path(:site), "robots.txt")
+    isfile(dst) && rm(dst)
+    io = IOBuffer()
+    globvar("generate_sitemap") && println(io, """
+Sitemap: $(joinpath(globvar(:website_url), "sitemap.xml"))
```
tlienart, line 20: you can indent this, it won't affect how it's written as it's the first line of a multi-line string. Here's an example:

```julia
print("""
    Hello this is the first line
    this is the second line""")
```

src/manager/robots_generator.jl
```diff
@@ -0,0 +1,28 @@
+# Sitemap: $(joinpath(path(:site), "sitemap.xml"))
+#
+# User-agent: *
+# Disallow:
+#
+# TODO
+# allow {{robots_exclude}} for a HTML page to add it to disallow
```
tlienart, line 7: I think it makes more sense to make this a global parameter where you specify paths to directories to exclude or files to exclude, same syntax as the `ignore` global variable

src/manager/robots_generator.jl
```diff
@@ -0,0 +1,28 @@
...
+    dst = joinpath(path(:site), "robots.txt")
+    isfile(dst) && rm(dst)
+    io = IOBuffer()
+    globvar("generate_sitemap") && println(io, """
+Sitemap: $(joinpath(globvar(:website_url), "sitemap.xml"))
+    """)
+    println(io, """
+User-agent: *
+Disallow:
+    """)
```
tlienart, line 25: same here

tlienart:
Nice thanks!

I think a simple test could be added where you pre-specify the base url by doing 

```
Franklin.set_var!(Franklin.GLOBAL_VARS, "website_url", "https://www.example.com")
```

and then call your function and check that it produces the right file in a temp folder

src/manager/robots_generator.jl
remi-garcia: At first I used an indentation but, for me, that does indent:
```Julia
julia> println("""
           User-agent: *
           Disallow:
       """)
    User-agent: *
    Disallow:


julia> println("""
       User-agent: *
       Disallow:
       """)
User-agent: *
Disallow:
```

tlienart: you need the final `"""` to be on the same level as the indentation or at the end

```julia
julia> println("""
               User-agent: *
               Disallow:
               """)
User-agent: *
Disallow:
```

src/manager/robots_generator.jl
remi-garcia: That would require to almost write a full `robots.txt` by hand. For folders I totally agree, yet for individual files a `@def disallow_robots = true` or similar seems simpler to use.

tlienart: that makes sense yes, I guess there's a choice as to how many such "disallow" a user would want. But I think it's good that what you're suggesting would also be similar to the `sitemap_exclude` which is also on a page level. Carry on :+1:

tlienart:
Hello Remi! would you like a hand with this or are you still meaning to do this when you have time? thanks!

remi-garcia:
Hello! I should have time today to finish, hopefully. If not I'll make time tomorrow for that. I think everything is almost ready for a push, and a review, but I had to stop in the middle of the work last Saturday and had no time to go back to it.

tlienart:
no worries, thanks!

remi-garcia:
TODO before merge:
- [ ] Add tests
- [ ] Naming! I'm not sure `robots_disallow` or `robots_disallow_this_page` is as nice as I first though 😞 
- [ ] Add info about robots in docs

TODO
- [ ] Add info about robots in FranklinTemplate.jl so `newsite()` prepare `@def nicenamefordisallowingrobots = []`

I think that's all

tlienart:
naming is always hard but in this case I think `robots_disallow` is just fine. Just make the docstrings for dummies and (ideally)  extend the docs `/docs/workflow/sitemap-rss` 

Thanks so much!

remi-garcia:
Well, rebase happened... Sorry, that makes this PR a bit fuzzy. Do you want to close it and to reopen a cleaner one?

PR #673, opened 2020-10-25
Use a ref for the highlight.js library to allow users to swap it out.

fredrikekre:
Used it locally so might as well push it upstream.


tlienart:
Thanks!

PR #676, opened 2020-10-27
Add huijzer.xyz to README

rikhuijzer:
Also removed some dots, since most of the lines did not end with a dot.


tlienart:
thanks!

PR #682, opened 2020-10-31
Sort RSS items by pubDate, fixes #681.

### LINKED ISSUE #681: feed.xml should be sorted by pubDate
The Feedbro plugin to FF only stores the X latest posts, and it seems to take the X first items in the XML file rather than parsing it all and sorting by date. This seems to suggest that the XML file should be sorted with newest entries at the top.

fredrikekre:


tlienart:
Great!

PR #686, opened 2020-11-01
Omit index.html from URLs in feed.xml.

fredrikekre:
Is this OK or does it have to include `index.html`? I could not find anything online.

### DIFF
--- a/src/manager/rss_generator.jl
+++ b/src/manager/rss_generator.jl
@@ -61,7 +61,7 @@ $SIGNATURES
 Create an `RSSItem` out of the provided fields defined in the page vars.
 """
 function add_rss_item()
-    link  = url_curpage()
+    link  = get_url(url_curpage())
     title = jor("rss_title", "title")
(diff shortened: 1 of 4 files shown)


fredrikekre:
I guess the same could be done for `sitemap.xml`?

tlienart:
I can't imagine it's necessary (though, like you, I didn't know and went for the conservative route but that's probably not needed), I'll merge this without releasing,  test it a bit with a couple of RSS readers and we can revert if there's something fishy going on.

Thanks for all the contribs!

fredrikekre:
On second though; perhaps it would be enough to prettify the URL here: https://github.com/tlienart/Franklin.jl/blob/41289c895f0cf3d0c65f35061d69558e44b02972/src/manager/rss_generator.jl#L188 since that is the only thing a human would see/click on. Perhaps that is better?

tlienart:
Yes that's true

PR #687, opened 2020-11-02
tag feeds: fix pretty URL and handle tags with special characters

fredrikekre:


fredrikekre:
Hmm. Actually, let me check some more stuff before merging.

fredrikekre:
Alright, so since the feed reader will most likely use the `<link>` element to generate a link for the user I think it is nice to keep them all "pretty". I don't think it should cause any problems.

tlienart:
cool! thanks Fredrik

PR #691, opened 2020-11-03
Remove sandbox module information in \show

### LINKED ISSUE #688: (demo) add note on clipboard.js
#680 for the steps. 

also this uses clipboard 1.4 or something, current version is 2.0; this might not matter at all but might be worth pointing it out (it probably doesn't matter much). 

Maybe a hover/image might be nice, something like https://doc.rust-lang.org/stable/book/ch07-02-defining-modules-to-control-scope-and-privacy.html but that could easily be done by the user

note: https://github.com/JuliaDocs/Documenter.jl/pull/1456

### LINKED ISSUE #690: Don't show module information in \show or allow eval in Main
As reported on Slack

````
```julia:tedious
struct T; v::Int; end
[
    Dict(:a => T(1)),
    Dict(:b => T(2)),
]
```
\show{tedious}
````

will show something like

```
2-element Vector{Dict{Symbol, Main.FD_SANDBOX_3080104478635972045.T}}:
 Dict(:a => Main.FD_SANDBOX_3080104478635972045.T(1))
 Dict(:b => Main.FD_SANDBOX_3080104478635972045.T(2))
```

instead of the REPL:

```
julia> [
           Dict(:a => T(1)),
           Dict(:b => T(2)),
       ]
2-element Vector{Dict{Symbol, T}}:
 Dict(:a => T(1))
 Dict(:b => T(2))
```

There could be two approaches:

1. allow running the code of a page in `Main`; possibly this could be indicated by a page var toggle, I'm reluctant to allow this as it would require disabling caching to avoid clashes between pages; this would make the overall build of a site slower potentially
1. modify the `\show` to replace `Main.FD_XXX.`

the second option seems to me to be a lot simpler and preferable

### weave note

> essentially we just use include_string and allow uses to explicitly specify the context module as an argument of entry methods:
https://github.com/JunoLab/Weave.jl/blob/c9fc740d72dbbd2f0b950f98efedc5ab71f17e3e/src/run.jl#L218

### PR DESCRIPTION
**Pull request by tlienart**
Removes `Main.FD_SANDBOX_....` when using `\show` of a code output, this leads to shown output as if they were evaluated in the REPL:

Screenshot 2020-11-03 at 11 11 48

What this PR does is simply to replace `Main.FD_SANDBOX_...` in output, this, I think, should generally be satisfactory (even if a user defines a module within a code block).

cc: @aviatesk who kindly reported this.

closes #690, also closes #688

### DIFF
--- a/src/converter/latex/io.jl
+++ b/src/converter/latex/io.jl
@@ -1,4 +1,6 @@
 """
+    \\input{qualifier}{rpath}
+
 Resolve a command of the form `\\input{qualifier}{rpath}` where
 
 * `qualifier` indicates what should be inserted (either a code block or a
@@ -72,6 +74,8 @@ end
 
 
 """
+    \\output{...}
+
 Return the output of a code. Possibly including the result of the code
 (`res=true`) and possibly reprocessing the whole (`reproc=true`).
 At most one will be true. See [`lx_show`](@ref).
@@ -95,7 +99,7 @@ function lx_output(lxc::LxCom, lxd::Vector{LxDef};
             if !isempty(output) && !endswith(output, "\n")
                 output *= "\n"
             end
-            output *= result
+            output *= replace(result, r"Main\.FD_SANDBOX_[0-9]+\." => "")
         end
     end
     # should it be reprocessed ?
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: src/converter/latex/io.jl [hunk 1]
```diff
@@ -95,7 +99,7 @@ function lx_output(lxc::LxCom, lxd::Vector{LxDef};
             if !isempty(output) && !endswith(output, "\n")
                 output *= "\n"
             end
-            output *= result
+            output *= replace(result, r"Main\.FD_SANDBOX_[0-9]+\." => "")
```
tlienart, line 102: @aviatesk this is the only change

fredrikekre:
I believe this is what Documenter does too.

src/converter/latex/io.jl
aviatesk: yeah string-based manipulation'd be the simplest fix. thanks !

PR #699, opened 2020-11-03
follow up #698

aviatesk:
We actually don't need to special case them ;)


tlienart:
Ah great! glad to have you on board thanks for the PR!

PR #702, opened 2020-11-05
Fix some links in FranklinDemo

tomohiro-soejima:
Some links in FranklinDemo were pointing to the archived FranklinFAQ, so I fixed them to point to the correct destinations.


tlienart:
thanks!

PR #732, opened 2020-12-13
Introduce parameters with spaces

### LINKED ISSUE #731: Allow Spaces in fun calls arguments
First of all, since I am switching from liquify/ruby to a Julia based generator: This is much more fun to build!

To my current problem: I have a scenario where my variable I'd like to pass to an `hfun_` contains a space. How can I achieve this?

For context: I have a `names.yml` that stores metadata for people (a full name, a short name, maybe an orcid or a link or a GitHub username or somethings like that – I can surely show a demo when I am finished). And in the ruby variant I used the Full name as a key in the yaml. To be precise: The first known full name (in case someone marries / changes their name).

So my function `fun_name` takes 2 arguments: the person to display and the format to use, i.e. I would call

```
{{name "Ronny Bergmann" link_fullname_fngithub}}
```
in order to display my name linking to my stored url and showing a GitHub icon in `<sup>` linking to my GitHub (if such data is stored in above mentioned `.yaml` file). However currently this ends up as 3 parameters
```
["\"Ronny", "Bergmann\"", "link_fullname_fngithub"]
```
How can I “escape” a space in there? Currently I trick this by storing the data under `RonnyBergmann`, i.e. I omit spaces, but it's extra effort and not nice in style, especially because I use that in another setting as identifier, too.

kellertuer:
This PR resolves #731, i.e. one can now pass parameters including spaces to an html function by encapsulating it in `"..."`.

To test define
````
function hfun_pcheck(params)
  return """<ul><li>$(join(params,"</li><li>"))</li></ul>"""
end
````

and run in an `.md` file for example
```
{{pcheck abc "abd ref h"}}
```
then you get the following unnumbered list.
Bildschirmfoto 2020-12-13 um 18 33 44

### DIFF
--- a/src/parser/html/blocks.jl
+++ b/src/parser/html/blocks.jl
@@ -59,7 +59,14 @@ function qualify_html_hblocks(blocks::Vector{OCBlock})::Vector{AbstractBlock}
             if isnothing(m.captures[2]) || isempty(strip(m.captures[2]))
                 ps = String[]
             else
-                ps = split(m.captures[2])
+                # split parameters such that values in "..." may contain spaces
+                splitregexp = r"([^\s\"']+)|\"([^\"]*)\""
+                ps = SubString.(
+                    m.captures[2],
+                    findall(splitregexp, m.captures[2]; overlap=false),
+                )
+                # strip leading and trailing "
+                ps = strip.(ps,Ref('"'))
             end
             qb[i] = HFun(β.ss, m.captures[1], ps)
             continue
--- a/src/parser/html/blocks.jl
+++ b/src/parser/html/blocks.jl
@@ -59,14 +59,8 @@ function qualify_html_hblocks(blocks::Vector{OCBlock})::Vector{AbstractBlock}
             if isnothing(m.captures[2]) || isempty(strip(m.captures[2]))
                 ps = String[]
             else
-                # split parameters such that values in "..." may contain spaces
-                splitregexp = r"([^\s\"']+)|\"([^\"]*)\""
-                ps = SubString.(
-                    m.captures[2],
-                    findall(splitregexp, m.captures[2]; overlap=false),
-                )
-                # strip leading and trailing "
-                ps = strip.(ps,Ref('"'))
+                # merge \"abc def\" (issue #731)
+                ps = split_hfun_parameters(m.captures[2])
             end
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Comment by tlienart** [comment]
Thanks a lot for this, would you mind the following small bit of refactoring?

```Julia
else
    # merge \"abc def\" (issue #731)
    ps = split_hfun_parameters(m.captures[2])
end
```

in the same file, at the end,

```julia
function split_hfun_parameters(s)
    matches = [e.match for e in eachmatch(HFUN_PARAMS_PAT, s; overlap=false)]
    return strip.(matches, '\"')
end
```

in `src/regexes`

```
"""
    HFUN_PARAMS_PAT

Splits the parameter string of a hfun based on whitespaces but allowing single quoted strings
with whitespace not to be split.
"""
const HFUN_PARAMS_PAT = r"([^\s\"']+)|\"([^\"]*)\""
```

you could then add a simple test in `test/converter/html/html2.jl` at the end:

```
# https://github.com/tlienart/Franklin.jl/issues/731
@testset "hfun strings" begin
    a = Franklin.split_hfun_parameters(strip("""
        "Hello to you" my "julia friends" and "all the others"
        """))
    @test a[1] == "Hello to you"
    @test a[2] == "my"
    @test a[3] == "julia friends"
    @test a[4] == "and"
    @test a[5] == "all the others"
end
```

Thanks a lot for looking into this, I think it's quite nice!

tlienart:
PS: quite unrelated but if you want to avoid your list looking too widely spaced because the Julia Base Markdown module adds `<p>` in every `<li>`, an easy way is to add `li > p {margin: 0}` in your CSS.

kellertuer:
Ah I was already wondering about all the wild `<p>`s apprearing. Thanks for the fix. Will add your changes in a minute (just doing a little larger test in my project – it really makes my code and especially my data much much nicer.

kellertuer:
You can, by the way, group `@testset`s into `@testset`s (even one global one) to have reduced output when running tests. But locally all tests are running here.

tlienart:
> You can, by the way, group `@testset`s into `@testset`s (even one global one) to have reduced output when running tests. But locally all tests are running here.

thanks yes I know but I prefer the full output :) thanks for the rest, will patch-release this

PR #748, opened 2020-12-28
Apply CDATASection to title of rss

5ebec:
In `feed.xml`, the description has a CDATA section, but the title does not. I propose to apply the CDATA section to the title as well.
This change will allow us to use `&`, `<` for article titles.

Thank you for developing such a great package!


tlienart:
Cool thank you, this seems useful! could you fix the RSS tests? otherwise I'll do this a bit later. Thanks!

5ebec:
I'm sorry, I'll fix it right now.

5ebec:
The current tests passed, but do we need to change `test/manager/rss.jl`?

For example, add the following after line 64 of it.
```julia
    @test occursin("<title><![CDATA[title 2]]></title>", fc)
    @test occursin("<title><![CDATA[title 1]]></title>", fc)
````

tlienart:
Don't hesitate to add tests, possibly adding comments explaining what they are, thanks! I'll check all this soon and merge, thanks again for contributing 😊

5ebec:
Thank you very much for your kind reply.
I have added tests for the RSS title.

tlienart:
Fantastic,  thank  you!

PR #753, opened 2021-01-06
convert_md: less diversity, better inference

timholy:
One of the biggest compile costs in Franklin is `convert_md`.
This standarizes the argument types to reduce the number of times
it needs to be compiled (xref https://timholy.github.io/SnoopCompile.jl/stable/snoopi_deep_analysis/#Argument-standardization)
and improves the inferrability for `lxdefs` and a couple of
other internal variables.

### DIFF
--- a/src/converter/markdown/md.jl
+++ b/src/converter/markdown/md.jl
@@ -20,8 +20,8 @@ Returns the html string as well as a dictionary of page variables.
 * `has_mddefs=true`:   a bool indicating whether to look for definitions of
                         page variables
 """
-function convert_md(mds::AbstractString,
-                    pre_lxdefs=collect(values(GLOBAL_LXDEFS));
+function convert_md(mds::String,
+                    pre_lxdefs::Vector{LxDef}=collect(values(GLOBAL_LXDEFS));
                     isrecursive::Bool=false,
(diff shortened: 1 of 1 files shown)


src/converter/markdown/md.jl
```diff
@@ -105,7 +103,7 @@ function convert_md(mds::AbstractString,
     #>> b. if any lxdefs are given in the context, merge them. `pastdef` specifies
     # that the definitions appeared "earlier"
     lprelx = length(pre_lxdefs)
-    (lprelx > 0) && (lxdefs = cat(pastdef.(pre_lxdefs), lxdefs, dims=1))
+    (lprelx > 0) && (lxdefs = vcat(LxDef[pastdef(def) for def in pre_lxdefs], lxdefs))
```
timholy, line 106: Because `LxDef` is a parametric type and `pre_lxdefs` is a `Vector{LxDef}` with no specified parameter, inference fails for lots of things including the broadcasting infrastructure. This avoids the need for inference to try to figure out what container type to use (https://timholy.github.io/SnoopCompile.jl/stable/snoopi_deep_analysis/#Declaring-container-types), which presumably saves time as well as achieving a better answer (inference decided it was `::Any` prior to this change).

PR #786, opened 2021-02-12
Support verifying relative internal links

mbauman:
Not sure how your test suite works or how to test this, but it works locally with an imported static site that very liberally used relative links.


tlienart:
lgtm, thanks Matt!

PR #802, opened 2021-03-18
a couple of missed _html_parts

kescobo:
Found these while trying to set up a new site.

There are also two instances of `_html_parts` remaining in `docs/_libs/lunr/lunr_index.js`, but I didn't want to edit those by hand since I've no idea how it's generated.


tlienart:
Ah thanks that's great!

PR #817, opened 2021-05-12
Add option to pass default value to pagevar

kescobo:
Forgive the presumption of an unasked for PR, but it seemed easier to just start this and if you don't want it, no problem to close.

I am finding myself regularly using formulations like 

```
function default_pagevar(path, var, default)
    val = pagevar(path, var)
    return isnothing(val) ? default : val
end
```
and I thought it might be nice to have that as an option in `pagevar` itself (and possibly `locvar` too?). This just adds an optional third argument with `nothing` as the default. It shouldn't change any existing code, but is a nice little convenience.

I couldn't find an obvious place to put new tests, but let me know where you'd like them, and if any other functions could use the same option and I'll add them in.

### DIFF
--- a/docs/syntax/utils.md
+++ b/docs/syntax/utils.md
@@ -84,7 +84,7 @@ function hfun_bar(params)
 end
 ```
 
-Note that all functions defined in `utils.jl` can call `locvar(name)` and `globvar(name)` to retrieve the value associated with a local or global page variable by its name; for instance `locvar("author")`.
+Note that all functions defined in `utils.jl` can call `locvar(name)` and `globvar(name)` to retrieve the value associated with a local or global page variable by its name; for instance `locvar("author")`. You may optionally pass a `default` argument that will be returned instead of `nothing` if the variable does not exist, eg `locvar(name; default="Not named")`.
 
 ## "LaTeX" functions (`lx_*`)
 
--- a/src/utils/vars.jl
+++ b/src/utils/vars.jl
@@ -219,7 +219,7 @@ The keys don't have the file extension so `"blog/pg1" => PageVars`.
 const ALL_PAGE_VARS = LittleDict{String,PageVars}()
 
 """
-    pagevar(rpath, name)
+    pagevar(rpath, name, default)
 
 Convenience function to get the value associated with a var available to a page
 corresponding to `rpath`. So for instance if `blog/index.md` has `@def var = 0`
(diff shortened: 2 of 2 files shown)


tlienart:
this makes sense to me and thanks!

A couple of requests:

* as you suggest, please add this to `locvar, globvar, pagevar` (i.e. not separate functions like `default_pagevar` but I imagine this is just your example for clarity)
* please make it a named argument so `default=...`
* (nitpicking) please use `ifelse` instead of `?` i.e. `ifelse(val === nothing, default, val)`

thanks!

kescobo:
> (nitpicking) please use ifelse instead of ? i.e. ifelse(val === nothing, default, val)

in this PR, at least for `pagevar`, I just changed `haskey(ALL_PAGE_VARS[rpath], name) || return nothing` to `... || return default` - is that OK? I'll look at the other functions now to see if they have similar constructions, if not, I'll use `ifelse` as you suggest :+1:

kescobo:
> as you suggest, please add this to locvar, globvar, pagevar (i.e. not separate functions like default_pagevar but I imagine this is just your example for clarity)

Yep, that's what I intended :heavy_check_mark: 

> please make it a named argument so default=...

:heavy_check_mark: 

> please adjust the docs :yum: 

Did you mean just docstrings? If so, :heavy_check_mark: 

> (nitpicking) please use ifelse instead of ? i.e. ifelse(val === nothing, default, val)

Ha! You initially used ternary expression in `locvar` and `globvar` :stuck_out_tongue:. For the later, `get` already exists for dictionaries to provide a default value, but since the others need to access the `.first` field, I went ahead and did the `ifelse` thing. For `pagevar`, it originally had the short-circuit version I mentioned above, but I changed it to be `ifelse` as well for consistency.

If there's anything else, please let me know!

kescobo:
ack - getting some testing failures... I introduced a bug

kescobo:
OK, so that took me a while to debug... Turns out I can't use `ifelse()` for these. Check it out:

```
julia> d = Dict("test"=> "hello");

julia> foo(dict, key, default) = haskey(dict, key) ? dict[key] : default
foo (generic function with 1 method)

julia> bar(dict, key, default) = ifelse(haskey(dict, key), dict[key], default)
bar (generic function with 1 method)

julia> foo(d, "blah", "it works!")
"it works!"

julia> bar(d, "blah", "it errors!")
ERROR: KeyError: key "blah" not found
```

I guess because the `bar` version needs to evaluate `dict[key]`, where the ternary expression doesn't. If you have a different idea, let me know, but since the ternary expression was there already, maybe it's not such a bit deal?

Weirdly, `pagevar` did not error with the `ifelse` version, which leads me to think that the version where the key doesn't exist is not actually being tested at the moment :-)

tlienart:
yes this is all good, you need a `?` because the `[]` access assumes the key is there, I guess you figured it out but you can do:

```
true ? 0 : sqrt(-1)
```

but you can't do

```
ifelse(true, 0, sqrt(-1))
```

in the first case the erroneous expression is not seen/evaluated, in the second it is.

Thanks for all this. The last minor thing I'd ask is if you could modify this https://github.com/tlienart/Franklin.jl/blob/master/docs/syntax/utils.md to just mention that you can specify a default with `default=...` and that otherwise `nothing` will be returned if the variable doesn't exist. Thanks!

kescobo:
Done! I couldn't find a reference to `pagevar` in the docs directory though.

tlienart:
Thanks a lot Kevin!

PR #824, opened 2021-05-31
More robust trim_stacktrace

tkf:
This PR makes `trim_stacktrace` more robust. I've been seeing build failure due to unguarded `first` in `first(findfirst(STACKTRACE_TRIM_PAT, s))`. Arguably, since the exact text representation of the stack trace is not part of the stable Julia API, it is ideal if `trim_stacktrace` can handle arbitrary string. Also, since `trim_stacktrace` is called at the error reporting portion of the code, I think it must not fail at all cost (c.f., `@error "hello" error("oops")` does not throw). This is why I am suggesting a catch-all patch. Of course, if this is not appropriate, I'm open to other implementation strategies (e.g., explicit `nothing` check before `first`).

### DIFF
--- a/src/eval/run.jl
+++ b/src/eval/run.jl
@@ -28,11 +28,17 @@ $SIGNATURES
 
 Returns only the stack traces which are related to the user's code.
 This means removing stack traces pointing to Franklin's code.
+Return the string as-is if the format is unrecognized.
 """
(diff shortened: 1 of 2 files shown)


tlienart:
Thanks a lot, this is much appreciated!

PR #839, opened 2021-06-25
adding optional time taken per page

tlienart:
The first pass can take some time depending on the complexity of individual pages and the hfuns used. This allows to pass a `show_timings=true` to `serve` to get a timing per page on the first full pass something like:

```
[δt = 0.11s] 09-ridge-cv-2.md
[δt = 0.85s] 09-convex-optimisation-3.md
```

### DIFF
--- a/src/Franklin.jl
+++ b/src/Franklin.jl
@@ -84,6 +84,7 @@ const FD_ENV = LittleDict(
     :SILENT_MODE   => false,
     :QUIET_TEST    => false,
     :SHOW_WARNINGS => true,     # franklin-specific warnings
+    :SHOW_TIMINGS  => false,    # first pass, how much time spent on each pg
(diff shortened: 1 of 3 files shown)


fredrikekre:
Whats the reason for (i) having this `false` by default and (ii) having it configurable at all?

tlienart:
hmm this is mostly for debugging purposes on my side, the initial full pass should be:

* some small delay for precompilation
* time taken by code blocks if any needs to be evaluated
* negligible time taken to process the rest (~10ms per page)

On my current website the last point is not negligible (~1s for some pages) so I wanted to investigate what exactly was causing a bottleneck, this is part of the process.

The reason for not setting this to true by default is that it's verbose and not really actionable information for the user, a technical user might be interested in it but generally my assumption is that a user just wants stuff to not feel sluggish.

PR #844, opened 2021-07-04
Add note on `+++`'s

logankilpatrick:
I am using the minimal mistakes template and Franklin was mad at me since I put the `@def` in the `+++`'s.

### DIFF
--- a/docs/workflow/deploy.md
+++ b/docs/workflow/deploy.md
@@ -26,7 +26,7 @@ Start by creating an empty GitHub repository
 @@
 
 \warn{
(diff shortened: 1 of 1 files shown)


tlienart:
ah yes well it should either be

```
+++
...
prepath = ...
...
+++
```

or

```
@def ...
@def prepath = ...
@def ...
```

best not to mix and match. 

I'm happy to have a note on this but I'd rather it didn't mix the `@def` and the `+++`. Maybe one can simply say that you should define a variable `prepath` in your `config.md` and since all templates ship with a `config.md`, they can just copy the style of other variable definitions.

PR #858, opened 2021-08-02
Add "code-output" class to code-output blocks

### LINKED ISSUE #856: Can't style code input and output together with \show
The code and the output are wrapped in `<pre>` but there is no way to apply css differently if an input block and an output block follow each other. Maybe it would be good to give and input and an output class to the `<pre>` tags?

tlienart:
Now:

````
julia> s = """
       ```!
       1+1
       ```
       """ |> fd2html |> println
<pre><code class="language-julia">1&#43;1</code></pre>
<pre><code class="plaintext code-output">2</code></pre>
````

so same as before but there's the extra class `code-output` that allows separate styling of code output blocks. 

closes #856

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "Franklin"
 uuid = "713c75ef-9fc9-4b05-94a9-213340da978e"
 authors = ["Thibaut Lienart <tlienart@me.com>"]
-version = "0.10.45"
+version = "0.10.46"
 
 [deps]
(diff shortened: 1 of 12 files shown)


rikhuijzer:
Nice feature! Feel free to copy the CSS from Books.jl for output blocks :smile: From [the docs](https://rikhuijzer.github.io/Books.jl/function_code_blocks.html):

image

tlienart:
@rikhuijzer would you mind opening an issue on FT for this please?

PR #864, opened 2021-08-15
Add note for custom domain of project gh-page

zfengg:
After adding a custom domain `stdsem.me` to my **project** gh-page [zfengg/StdSem](https://github.com/zfengg/StdSem), I found the CSS didn't work in the first place and it looks terrible.

However, by setting the important global variable `@def prepath="" `, i.e., the empty string, the website [stdsem.me](https://stdsem.me) looks fine again. Hence I think this may be helpful for others who met the same problem.

Really appreciate `Franklin.jl` :+1: .


tlienart:
Thanks!

PR #872, opened 2021-08-22
Re-phrase the auto generation of the index js

logankilpatrick:
Just a quick phrasing change to clarify that the file is not supposed to be manually created.


tlienart:
Thanks!

PR #877, opened 2021-09-04
Add `hfun_requiredfill`

rikhuijzer:
Just a suggestion. Feel free to close this if you don't like the idea :slightly_smiling_face:

### DIFF
--- a/src/converter/html/functions.jl
+++ b/src/converter/html/functions.jl
@@ -79,7 +79,18 @@ function hfun_fill(params::Vector{String})::String
     end
     return repl
 end
+    
+"""
+    hfun_requiredfill(params::Vector{String})
 
+H-Function similar to `hfun_fill`, but this function throws an error if a field is not set.
+"""
+function hfun_requiredfill(params::Vector{String})::String
+    value = Franklin.hfun_fill(params)
+    field = params[1]
+    @assert(value != "", "Missing a value for the field $field")
+    return value
+end
 
 """
 $(SIGNATURES)


tlienart:
Thanks!

Two notes for discussion:

1. I think that now with the `fail_on_warning` option to serve, it would be a similar behaviour?
2. Generally I'd rather not have lots of new hfuns and lxfuns in Franklin but rather make those part of a side repo or a set of side repos like FranklinUtils so that maintenance is easier and decoupled

Wdyt?

rikhuijzer:
Yeah, that's what I thought. However, fill doesn't throw an error if the value is empty. It used to be like that but got changed somewhere between 0.42 and 0.52 (these numbers are by heart so might be wrong).

Would you rather have fill fail on empty or me opening a PR at FranklinUtils.jl?

tlienart:
Hmm probably a PR in FranklinUtils please

In the next major version I'll try to think with a bit more rigour about what should fail / warn / do nothing, but it'll end up being a bit breaking so I don't want to rush this.

If you have thoughts on that topic btw, please open an issue and let me know :-)

rikhuijzer:
> Hmm probably a PR in FranklinUtils please

You also seem to have some doubts and me too. Let's not do it. The functionality is reasonably trivial.

tlienart:
No I think the util is useful, my hesitation is that I'm preparing a monster release which, while it will try to preserve most of the Franklin surface as it is today, changes pretty much everything in the background and I was thinking about where contribs like yours should live. The thing is that FranklinUtils will also need to be revamped following that release as the way lxfuns are processed will be simplified (that of hfuns will remain pretty much identical). That's for the context :) 

Maybe once this has been done and the dust has settled, I'd like to discuss with you if you're interested how we could try to  encourage "plugins" for Franklin where people would load packages a bit like `.sty` files in LaTeX with a bunch of useful `hfun_*` and `lxfun_*` as well as utils to make it easy to write those.

PR #882, opened 2021-09-10
Update doc for mdstring support in Literate

luraess:
- Add mdstring doc
- Fix lower bound of Literate.jl compat


tlienart:
thanks a lot!

PR #916, opened 2021-10-31
Remove unsupported `force` kwarg from `FD_CAN_PRERENDER`

mcognetta:
The `force` kwarg is not present in the `FD_CAN_PRERENDER` definition and causes the build to fail when deploying a site.


tlienart:
thanks a lot!

PR #928, opened 2021-11-25
Use `\fig{}` to give a `PlotlyJS` plot.

### LINKED ISSUE #890: Feature request: display interactive figures given by Plots.jl with PlotlyJS backend just like fdplotly
When `PlotlyJS.jl` is used, a beautiful interactive plot can be display with the helper function `fdplotly`.
But I couldn't find a build-in method to show **interactive** `Plots.jl` figures with `plotlyjs()` backend just like when `PlotlyJS.jl` is used directly. (Don't know whether that is the case.)

So I have implemented a helper function myself. It works but I think it is not so good that several problem has to be solved.
```julia
function fdplotsplotlyjs(p; id="fdp"*Random.randstring('a':'z', 3),
  style="width:600px;height:350px")::Nothing
  print("""
    ~~~
    <div id="$id" style="$style"></div>
    <script>
      var fig = """)
  show(stdout, MIME("application/vnd.plotly.v1+json"), p);
  println("""; CONTAINER = document.getElementById('$id');
      Plotly.newPlot(CONTAINER, fig.data, fig.layout)
    </script>
    ~~~
    """) # a ";" is used to end var fig = ...;
end
```

It can be used like this in `Franklin.jl`
~~~julia
 ```julia:ex
using Plots
plotlyjs()
fdplotsplotlyjs(plot([1, 2])
```
\textoutput{ex}
~~~

Several Problem should be solved:
1. I think there should be a better way to transfer the plot to a JSON format but don't know how. The function [`json`](https://github.com/JuliaPlots/Plots.jl/blob/54a6518d59498fb24b5c25a90cd91ceadb7e26de/src/output.jl#L55) in `Plots.jl` directly save it to a file which I think is unnecessary.
2. I think it should be better to give a type declaration on `p`. It should be something like `p::Plots.Plot{Plots.PlotlyJSBackend}`. But when this function is placed in `src/manager/extra.jl`, it can't find `Plots`.
3. Give the function a better name.

### PR DESCRIPTION
**Pull request by ChasingZenith**
Now, one can use `\fig{}` to give a `PlotlyJS` plot.

When `\fig{}` receive a json file, it will regard it as a `plotlyjs` json file and plot it using a JavaScript function `PlotlyJS_json` defined in `src/_layout/head.html` (One should add it in it.)

Documentation is also updated.

I suggest deprecate `fdplotly`, since it will slow down page loading. Details are described [here](https://github.com/tlienart/Franklin.jl/issues/890#issuecomment-943965952).

Closes #890

### CODE REVIEW

**Comment by tlienart** [comment]
thanks a lot for investigating this, it looks good to me, will test a bit then merge

**Comment by tlienart** [comment]
hmm when working on your branch, `cd("docs"); serve()` (after copy pasting your JS into the head) I get:

Screenshot 2021-11-25 at 17 27 32

Could you please try this?

File: docs/extras/plotly.md [hunk 1]
```diff
@@ -19,7 +19,24 @@ Then create a variable `hasplotly` with default value `false`: in your `config.m
...
+    // Make the plot fit the screen responsively. See the documentation of plotly.js. https://plotly.com/javascript/responsive-fluid-layout/
+    if (typeof fig.config === 'undefined') { fig["config"]={} }
+    delete fig.layout.width
+    delete fig.layout.height
+    fig["layout"]["autosize"] = true
+    fig["config"]["autosizable"] = true
+    fig["config"]["responsive"] = true
+
+    Plotly.newPlot(CONTAINER, fig.data, fig.layout, fig.config);
+  </script>
```
tlienart, line 38: I think there's a missing `}` opened on line 25 but never closed?

docs/extras/plotly.md
```diff
@@ -111,3 +128,30 @@ plt    = plot(data, layout)
 fdplotly(json(plt)) # hide
 ```
 \textoutput{ex1}
+
+### Use `\fig{}`
```
tlienart, line 132: ``Using `\fig{...}` (recommended)``

docs/extras/plotly.md
```diff
@@ -111,3 +128,30 @@ plt    = plot(data, layout)
...
+Now you might use `\fig{}` to insert graph [just like normal](/syntax/markdown/#inserting_a_figure). This also work fine with `Plots.jl` and `PlotlyBase.jl`.
+
+**Note**: `\fig{}` will call the Javascript function `PlotlyJS_json` defined [above](#pre-requisites). You might customize the behavior by modifying it. Also make sure `@def hasplotly = true` is properly set.
```
tlienart, line 135: `by modifying the Javascript`.

docs/extras/plotly.md
tlienart: also please add this to `docs/_layout/head` so that the docs work

ChasingZenith:
Sorry for making this silly mistakes and wasting your time. I was not aware of that I can `cd("docs"); serve()` and only did some simple copying and pasting.

I can confirm all problems you pointed out are fixed. Some improvements are added to get consistent experience when using `Plots.jl`, `PlotlyJS.jl` and `PlotlyBase.jl`.

tlienart:
Thanks a lot for your contribution @ChasingZenith sorry for the delay I was out last week; this seems to work fine on my side and I'm happy to integrate it.

ChasingZenith:
Thank you! It is my honor to make this contribution.

Would you consider depercating `fdplotly`?

As an minimal example, the following extremely simple plot give a 6.9kB json with `PlotlyBase`, 12.4kB json with `PlotlyJS`. 
```julia
using PlotlyBase               #  6.9 kB json
# using PlotlyJS               # 12.4 kB
# using Plots;  plotlyjs();    #  2.1 kB (BTW, it is strange that PlotlyBase and PlotlyJS give larger json than Plots)
# using Plots;  plotly();      #  2.1 kB
p = Plot([0, 1]);  # use function `plot` when not using PlotlyBase
savejson(p,"2num.json")
```

The html file of a page is around 25kB, taking this page [franklinjl.org/code](https://franklinjl.org/code/) as an example.

Directly inserting the json file into the html, just like what `fdplotly` does can heavily increase loading time. Especially when some complicated plot were added, the html file can grow to 10MB or so.

ChasingZenith:
Thank you!

atiyabzafar:
Hey everyone. I know this has merged and the [Issue](https://github.com/tlienart/Franklin.jl/issues/890) has been resolved. Great work by @ChasingZenith and @tlienart for this. 

But I am facing an interesting problem here. I am serving my static site locally on a windows machine. I am using `\fig{}` to add a figure that is same as the example that you have in the documentation. 
This is what it looks like: https://imgur.com/a/6JTyPCJ

I did some troubleshooting and found that in the function `PlotlyJS_json = async (div, url)` url is the one that is defined by the location of the json file. Further according to my understanding, 

`savejson(p, joinpath(@OUTPUT, "plotlyex.json"))`

This saves the file by merging `@OUTPUT` variable which is the environment location and name of the json file to save the file there. So i checked the json file and the json file is being saved at the right place and is correct. Therefore, I decided to check the html that Franklin is generating from the markdown file and I found the _culprit_. 

```
<div id="fdpitp" style=""></div>
<script>
graphDiv = document.getElementById("fdpitp");
plotlyPromise = PlotlyJS_json(graphDiv, '/assets/Blog/test/code\output\plotlyex.json');
</script>
```

The location of the json file is messed up. Instead of forwardslash the location has backslash. `\fig{}` somehow is not able to convert the file location properly according to my understanding. I was not sure if this issue is serious enough for opening a new issue. Maybe I am making some big blunder and am just missing something. Help would be appreciated.

I am using conda on windows and have julia installed the usual way if that is important.

tlienart:
> EDIT2: **Is it possible that systempath and other path are different due to how windows define path v/s how a linux or mac would?** In here:

yes, windows is extremely annoying in this respect, and browsers use the unix style paths, so there are functions to go back and forth. Looks like it's missing here. 

Could you kindly try again but using `master` instead of the latest patch release? (i.e. commit 6077d23c), it should fix this. Assuming it does, I'll do a patch release with it. Thanks!

atiyabzafar:
Hey, Apologies for not being able to reply yesterday.

I did what was asked. Instead of actually using it on my own Franklin site, I served the docs from the cloned repo. So that you can compare results better.

This is what I get from the output:
```
...
→ evaluating code [ex2] in (extras\plotly.md)
rpath = "plotlyex"
path = "/assets/extras/plotly/code/plotlyex"
(fdir, fext) = ("/assets/extras/plotly/code/plotlyex", "")
(p1, p2) = ("/assets/extras/plotly/code", "plotlyex")
...
```
This is what I am seeing on the frontend:
image

Path should have `.json` as an extension right?

plotly.json is present in `/assets/extras/plotly/code/output`

tlienart:
Thanks for this work, this looks almost right. Can you try reverting what I did in the previous commit so:

```Julia
    if splitdir(p1)[2] != "output"
        for ext ∈ candext
            #candpath = joinpath(p1, "output", p2 * ext)
	    candpath= p1* "/output/" * p2*ext
		@show candpath
   # now candpath is in unix-style, so it should be split according to "/"
            syspath  = joinpath(PATHS[:site], split(candpath, "/")...)
            # with join path, it should now be correct irrelevant of the platform
		@show syspath
#		if Sys.iswindows()
#    			# syspath=replace(syspath,"/"=>"\\")
#	    		isfile(syspath) && return ext == ".json" ? html_plotly(candpath) : html_img(candpath, alt)
 #                       @show isfile(syspath)
#		else
            		isfile(syspath) && return ext == ".json" ? html_plotly(candpath) : html_img(candpath, alt)
#        	end
	end
    end

```

atiyabzafar:
Hey @tlienart 

It is working now. But you would need to do the same changes in other loop as well. 
The following code in `\syntax\markdown\` does not work anymore. 
```
@@small-img
  \fig{./output/test}
@@
```

But the code following it works. i.e.: 
```
@@small-img
  \fig{./test}
@@
```


Pasting the truncated output to show how the path is now being defined :

```
candpath = "/assets/extras/plotly/code/output/plotlyex.json"
syspath = "E:\\work\\Franklin.jl-master\\Franklin.jl-master\\docs\\__site\\assets\\extras\\plotly\\code\\output\\plotlyex.json"
isfile(syspath) = true
```

PR #934, opened 2021-12-11
Clarify rss metadata example syntax

mcognetta:
This clarifies that rss metadata must be preceded by `@def` since they are local page variables. The other page on rss includes the complete syntax, so this PR is to make it uniform.


tlienart:
Thank you!

PR #935, opened 2021-12-23
Fix tag's RSS head templating

McSinyx:
Currently a tag's RSS feed's head is not filled, e.g. https://cnx.srht.site/tag/fun/feed.xml


tlienart:
oops, I had seen this a while back and not acted on it, sorry! this looks great, thanks

PR #942, opened 2022-01-21
Small addition to the documentation regarding Plots.jl

davibarreira:
I was trying to use Franklin.jl with Plots.jl, and I kept getting erros in the GitHub Actions. The issue was that I needed to add an environment variable in the `Deploy.yml`. I got this info from @fredrikekre by asking in the Slack. Thus, I find it to be useful to be added to the docs.


tlienart:
Thanks!

davibarreira:
Hopefully I'll submit a new template "soon" :grin:

PR #950, opened 2022-03-09
Allow classes like a:b/c with @@

xukai92:
Tailwind CSS (https://tailwindcss.com/) is a perfect match for `@@` but it has a many class names like `hover:basis-1/2` that the parser failed to reconize due to pre-matured end-of-token.
This PR fixes this issue.

### DIFF
--- a/src/parser/tokens.jl
+++ b/src/parser/tokens.jl
@@ -253,7 +253,7 @@ In combination with `incrlook`, checks to see if we have something that looks
 like a @@div describing the opening of a div block. Triggering char is a first
 `@`.
 """
-is_div_open(i::Int, c::Char) = (i == 1 && return c == '@'; return α(c, ('-','_', ',', NUM_CHAR...)))
+is_div_open(i::Int, c::Char) = (i == 1 && return c == '@'; return α(c, ('-','_', ',', ':', '/', NUM_CHAR...)))
 
 """
 $(SIGNATURES)
--- a/test/parser/1-tokenize.jl
+++ b/test/parser/1-tokenize.jl
@@ -118,7 +118,7 @@ end
         B3
     C
     @@da
-        @@db
+        @@db,dc:dd/de
             E
         @@
     @@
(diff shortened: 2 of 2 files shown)


tlienart:
Nice one, thanks @xukai92, happy to add that, also this should still work fine with 
https://github.com/tlienart/Franklin.jl/blob/43e23a1bbca17fe13c028872e99fe2a0af714fc3/src/converter/markdown/blocks.jl#L39-L40 

so should be good. I'll add a note in the docs and will do a patch release with it. 

Ps: will be good to see what you do with tailwind, do you already have a repo?

tlienart:
Nice one, so you'll use the base `tailwind.min.css` and then use `purgecss` at a later stage? I'm experimenting with all this in another repo so we can share knowledge if you want. I'll keep an eye on your repo for sure :)

tlienart:
Ah no I see https://github.com/xukai92/beta.turing.ml/commit/51cf53527454bc8110d97dc66fa9f79b65406172 

I played a bit with tailwindcss but wasn't using their JS build, rather just using the full CSS for development, then using purgecss to trim it down to a few kb post build.

xukai92:
I think what you do is the recommanded way from Tailwind's offical website---I simply did the easiest thing to get started.
Can you set it up to purgecss when deployment?

tlienart:
So what I do (I personally prefer bootstrap but tried tailwind to check):

- have the full blown mega CSS file in _css (the tailwind one is something ridiculous like 3MB)
- do stuff locally until happy
- in the deploy script after the `optimize`, use NodeJS to install purgecss then call it

Note: with tailwind the purgecss default extractor doesn't work well, you have to change the regex a little bit to include `:`

This is the config file I used with the tailwind case: https://github.com/tlienart/GLEPlot.jl/blob/docs2/docs2/tailwind.config.js

xukai92:
That's super useful! Thanks for sharing this!

PR #968, opened 2022-07-23
fix clipboard script for highlight.js pre-render

johnnychen94:
```julia
using NodeJS; run(`$(npm_cmd()) install highlight.js`);
using Franklin; Franklin.optimize()
```

I'm not sure if this is the right fix (almost zero web-dev knowledges). But it works locally.


tlienart:
lgtm, thanks

PR #970, opened 2022-07-31
Fix mini typos in the docs

svilupp:
As part of the JuliaCon hackathon, I have run all the docs through Grammarly. In case of ambiguity, I have picked the more common spelling on the site or googled for the more variant.

I'm happy to withdraw this PR if it's not helpful.


tlienart:
woah this is great, thank you very much!

PR #975, opened 2022-09-08
remove unbound type parameter

nsajko:
Unbound type parameters often cause performance issues and run time dispatch.

Issue found using https://github.com/JuliaLang/julia/pull/46608


tlienart:
Nice that you found this, thanks!

PR #984, opened 2022-10-11
Make `div_content` settable from a page

### LINKED ISSUE #983: Unable to set `div_content` with local page variable
According to the docs, the global page variable `div_content` should be overwritten if set as local page variable, but this does not happen.

I'm guessing this can easily be fixed by using `locvar` instead of `globvar` in the `build_page` method on [src/manager/write_page.jl](../blob/6f24bef91a256731ca7a3609e4b777fe71594200/src/manager/write_page.jl)

The new `content_class` variable does also not support setting it from a local page.

stanvanlier:
Fixes #983

### DIFF
--- a/src/manager/write_page.jl
+++ b/src/manager/write_page.jl
@@ -9,11 +9,11 @@ Convenience function to assemble the html of a page out of its parts:
 function build_page(head, content, pgfoot, foot)
     # (legacy support) if div_content is offered explicitly, it takes
     # precedence, otherwise use defaults
-    dc = globvar("div_content")
+    dc = locvar("div_content")
     if isempty(dc)
-        content_tag   = globvar("content_tag")
-        content_class = globvar("content_class")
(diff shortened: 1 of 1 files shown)


tlienart:
Thanks!!

tlienart:
Ps: do you need a patch release with this or are you ok to work with master for now?

stanvanlier:
Yes would like to deploy using gitlab ci, or is that also possible with master?

tlienart:
It is, the only difference is that in the gitlab script you'll have to install a specific version with something like

```
using Pkg
Pkg.add(PackageSpec(name="Franklin", rev="master"))
using Franklin
Franklin.optimize(...)
```

as opposed to just putting Franklin in the Project toml.

Happy to do a patch release as well, might do that within 24h anyway

stanvanlier:
Ah ok thanks! I'll use master for now then

PR #1006, opened 2023-04-13
Add demo to render Dataframe as HTML table

theogf:
This adds a small section in the demo section about rendering tables as HTML.


theogf:
It looks like the math rendering does not work so great...
`"\$\\frac{1}{2}\$"` get rendered as `$\\sqrt{2\\pi}$\n`

PR #1013, opened 2023-04-23
Clarify which files are ignored or copied.

fkastner:


fkastner:
Ok, I changed the link to a permalink. This might also become obsolete but I feel it is better to have an old link that still points "in the right direction" than no link.

PR #1019, opened 2023-04-24
update compathelper

fkastner:
This changes compathelper to only run every friday at 12am or if manually triggered (which is what the workflow_dispatch trigger does).

There also one functional change. It now runs always on the latest julia version instead of 1.2.0. I'm not sure if this has any (bad) consequences.

### DIFF
--- a/.github/workflows/CompatHelper.yml
+++ b/.github/workflows/CompatHelper.yml
@@ -2,25 +2,20 @@ name: CompatHelper
 
 on:
   schedule:
-    - cron: '00 * * * *'
-  issues:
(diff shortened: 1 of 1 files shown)


fkastner:
Turns out that CompatHelper doesn't even work currently. It runs every hour and checks which dependencies are outdated but it doesn't open any PRs.
To fix this it needs a new SSH key. Could you do this: [CompatHelper#Creating an SSS Key](https://juliaregistries.github.io/CompatHelper.jl/dev/#Creating-SSH-Key)?

PR #1021, opened 2023-04-26
fix tikz demo

fkastner:
This should fix the tikz demo.
But I commented out a part of the newest demo which, I think, breaks the later demos.


fkastner:
Another idea would be to migrate [TikzPictures.jl](https://github.com/JuliaTeX/TikzPictures.jl) to the [tectonic.jll](https://github.com/JuliaBinaryWrappers/tectonic_jll.jl) which nowadays seems to have good support.

Also: if the demo works on Github Pages then it's a valid demo for many people (those not with an M1 Mac). In my opinion, that still makes it very useful.

fkastner:
Does this now work for you?

fkastner:
Ok, this should be good now.

PR #1026, opened 2023-05-09
Sitemap fix for 404.md

### LINKED ISSUE #1011: wrong sitemap for `404.html`
Another sitemap issue:
the `404.html` appears as `<loc>https://tlienart.github.io/FranklinTemplates.jl/404/index.html</loc>`.
I think this is because the file is special cased in `form_output_path`
https://github.com/tlienart/Franklin.jl/blob/caa1518240ccff7f629d077d72eb65a09515b64f/src/manager/dir_utils.jl#L29-L45
and the sitemap generator does not know about this.

I think the most elegant way to handle this would probably be to add the file to `keep_path` instead of the current special handling. But that would be a breaking change.

tlienart:
Closes #1011 

@fkastner I'd appreciate a double check here; what I did:

```
julia> newsite(".")
julia> # added generate_sitemap = true in config.md
julia> serve(single=true)
shell> cat __site/sitemap.xml
(...)
<url>
    <loc>https://tlienart.github.io/FranklinTemplates.jl/404.html</loc>
    <lastmod>2023-05-09</lastmod>
    <changefreq>monthly</changefreq>
    <priority>0.5</priority>
</url>
(...)
```

thanks for the report!

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "Franklin"
 uuid = "713c75ef-9fc9-4b05-94a9-213340da978e"
 authors = ["Thibaut Lienart <tlienart@me.com>"]
-version = "0.10.82"
+version = "0.10.83"
 
 [deps]
 Dates = "ade2ca70-3891-5945-98fb-dc099432e06a"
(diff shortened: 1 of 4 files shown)


tlienart:
yeah that's a good point, I'll remove it.

PR #1030, opened 2023-06-05
return "python" when pyenv is used

### LINKED ISSUE #1028: "py -3" command does not work when pyenv is used.
Hi! Thank you for the great package!
I have one question related to windows user.

If windows user installed Python with pyenv, I think Python would be launched by "python" command.  
Therefore, "py -3" will cause an error in FD_HAS_PY3().
How about returning "python" when ENV["PYENV"] exists?

https://github.com/tlienart/Franklin.jl/blob/5c8f5e1af7d1279f5aad9e1437f7d883f4302307/src/build.jl#L1-L11

SosUts:
Closes #1028

### DIFF
--- a/src/build.jl
+++ b/src/build.jl
@@ -34,12 +34,7 @@ end
 # highligh.js library; can be overridden from the outside which is useful for testing
 const HIGHLIGHTJS = Ref{String}("highlight.js")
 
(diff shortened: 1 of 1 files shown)


tlienart:
Thanks!!

PR #1035, opened 2023-07-09
REPL mode for code blocks

### LINKED ISSUE #818: [Feature Request] Executed REPL block
## Request

Transferred discussion from slack thread: it would be great to have a feature for code output analogous to to Documenter's [`@repl` blocks](https://juliadocs.github.io/Documenter.jl/stable/man/syntax/#@repl-block), where something like

````
```@repl
foo = 1
foo + 2
```
````
is rendered like
````
```julia-repl
julia> foo = 1
1

julia> foo + 2
3
```
````

## Proposal

As suggested by @tlienart on slack 

> well we already have
> ````
> ```!
> 1+1
> ```
> ````
> to mean "run it, show the code + show the output here" i.e.: julia code block with <1+1> and output code block with <2>
> We could identically have
> ````
> ```>
> 1+1
> ```
> ````
> to mean "run it + show the code+output in REPL style" i.e.: julia-repl code block with
> ```
> julia> 1+1
> 2
> ```
> ````

which I think would be great, presuming we can do multi-line blocks, and connect blocks, eg

````
```>:foo
bar = 1
println(bar)
```

some explanation text

```>:foo
bar + 2
```
````

renders as

````
```julia-repl
julia> bar = 1
1

julia> println(bar + 1)
2
```

some explanation text

```julia-repl
julia> bar + 2
3
```
````

## Syntax

> also please chip in syntax wise, so if it should be
> ````
> ```> or ```? or ```$ or ...
> ````

I would expect ```` ```?``` ```` to be a help REPL, and ```` ```$``` ```` to be shell, so I wouldn't use those. Since `!` already has precedent, I think `>` would make a lot of sense, though ```` ```julia>```` would be the most explicit, I think. I also like slightly more explicit, since it will be eventially be easier to edit grammars to enable syntax highlighting for these blocks in markdown files.

### PR DESCRIPTION
**Pull request by tlienart**
ongoing todo list here: https://github.com/tlienart/Franklin.jl/pull/1035#issuecomment-1628923119

---

Can now do

````
```>
x = 5
y = x * 2
```
some explanation
```>
z = 2y
```
````

and get the following html

```
<pre><code class="language-julia-repl">
julia> x &#61; 5
5
julia> y &#61; x * 2
10
</code></pre>

<p>some explanation</p>
<pre><code class="language-julia-repl">
julia> z &#61; 2y
20
</code></pre>
```

(note `&#61` is `=`)

closes #818 

cc @gdalle , @kescobo 

Screenshot 2023-07-09 at 22 25 49

### DIFF
--- a/docs/code/index.md
+++ b/docs/code/index.md
@@ -328,11 +328,12 @@ gives
 st
 ```
 
-_Shell mode_ :
+_Shell mode_ : (**note**: in a multi-line setting, each line is assumed to be a separate command)
 
 ````
 ```;
 echo "hello!"
+date
 ```
 ````
 
@@ -340,6 +341,7 @@ gives
 
 ```;
 echo "hello!"
+date
 ```
 
 ~~~
--- a/docs/code/index.md
+++ b/docs/code/index.md
@@ -312,6 +312,78 @@ x = 5
 y = x^2
 ```
 
+**Shell, Pkg, Help**, these modes are also experimentally supported:
+
+_Pkg mode_ :
+
+````
+```]
+st
+```
+````
+
+gives
+
+```]
+st
+```
+
+_Shell mode_ :
+
+````
+```;
+echo "hello!"
+```
+````
+
+gives 
+
+```;
+echo "hello!"
+```
+
+~~~
+<style>
+.julia-help {
+    background-color: #fffee0;
+    padding: 10px;
+    font-style: italic;
+}
+.julia-help h1,h2,h3 {
+    font-size: 1em;
+    font-weight: 500;
+}
+</style>
+~~~
+
+_Help mode_ :
+
+````
+```?
+im
+```
+````
+
+```?
+im
+```
+
+**Note**: for the `help` mode above, the output is HTML corresponding to the julia
+docs, it's wrapped in a `julia-help` div which you should style, the above style
+for instance corresponds to the following CSS:
+
+```css
+.julia-help {
+    background-color: #fffee0;
+    padding: 10px;
+    font-style: italic;
+}
+.julia-help h1,h2,h3 {
+    font-size: 1em;
+    font-weight: 500;
+}
+```
+
 ### Troubleshooting
 
 A few things can go wrong when attempting to use and evaluate code blocks.
--- a/docs/code/index.md
+++ b/docs/code/index.md
@@ -358,7 +358,7 @@ date
 </style>
 ~~~
 
-_Help mode_ :
+_Help mode_ : (**note**: only single line cell blocks will work properly)
 
 ````
 ```?
--- a/docs/code/index.md
+++ b/docs/code/index.md
@@ -277,6 +277,41 @@ savefig(joinpath(@OUTPUT, "sinc.svg")) # hide
 
 **Note**: If you wish to use `Plots.jl` and deploy to GitHub pages, you will need to modify the `.github/workflows/Deploy.yml` by adding `env: GKSwstype: "100"` before the ` - name: Build and Deploy` line. [Here](https://github.com/storopoli/Bayesian-Julia/blob/master/.github/workflows/Deploy.yml) is an example.
 
+### Auto and REPL mode
+
+You can use `!` and `>` to indicate respectively a code that should be run automatically
+and the output appended immediately after, or the same but with a REPL-style display:
+
+````plaintext
+```!
+x = 5
+y = x^2
+```
+````
+
+for instance gives:
+
+```!
+x = 5
+y = x^2
+```
+
+In a similar way:
+
+````plaintext
+```>
+x = 5
+y = x^2
+```
+````
+
+gives
+
+```>
+x = 5
+y = x^2
+```
+
 ### Troubleshooting
 
 A few things can go wrong when attempting to use and evaluate code blocks.
--- a/src/converter/markdown/blocks.jl
+++ b/src/converter/markdown/blocks.jl
@@ -14,8 +14,11 @@ function convert_block(β::AbstractBlock, lxdefs::Vector{LxDef})::AS
 
     βn == :CODE_INLINE     && return html_code_inline(stent(β) |> htmlesc)
     βn == :CODE_BLOCK_LANG && return resolve_code_block(β.ss)
-    βn == :CODE_BLOCK!     && return resolve_code_block(β.ss, shortcut=true)
-    βn == :CODE_REPL       && return resolve_code_block(β.ss, repl=true)
+    βn == :CODE_BLOCK!     && return resolve_code_block(β.ss; shortcut=true)
+    βn == :CODE_REPL       && return resolve_code_block(β.ss; repl=true)
+    βn == :CODE_PKG        && return resolve_code_block(β.ss; pkg=true)
+    βn == :CODE_HELP       && return resolve_code_block(β.ss; help=true)
+    βn == :CODE_SHELL      && return resolve_code_block(β.ss; shell=true)
     βn == :CODE_BLOCK      && return html_code(stent(β), "{{fill lang}}")
     βn == :CODE_BLOCK_IND  && return convert_indented_code_block(β.ss)
 
--- a/src/converter/markdown/blocks.jl
+++ b/src/converter/markdown/blocks.jl
@@ -15,6 +15,7 @@ function convert_block(β::AbstractBlock, lxdefs::Vector{LxDef})::AS
     βn == :CODE_INLINE     && return html_code_inline(stent(β) |> htmlesc)
     βn == :CODE_BLOCK_LANG && return resolve_code_block(β.ss)
     βn == :CODE_BLOCK!     && return resolve_code_block(β.ss, shortcut=true)
+    βn == :CODE_REPL       && return resolve_code_block(β.ss, repl=true)
     βn == :CODE_BLOCK      && return html_code(stent(β), "{{fill lang}}")
     βn == :CODE_BLOCK_IND  && return convert_indented_code_block(β.ss)
 
--- a/src/eval/codeblock.jl
+++ b/src/eval/codeblock.jl
@@ -189,25 +189,19 @@ function resolve_code_block(
             end
 
         elseif pkg
-            # NOTE: this is very elementary, doesn't consider any `--` arguments
-            # etc.
-            # assume single line, first thing = what, rest = args
-            verb, args... = split(code)
-            if verb == "st"
-                verb = "status"
-            elseif verb == "remove"
-                verb = "rm"
+            for line in split(code, '\n', keepempty=false)
+                a = tempname()
+                open(a, "w") do outf
+                    redirect_stdout(outf) do
+                        redirect_stderr(outf) do
+                            Pkg.REPLMode.pkgstr(string(line))
+                        end
+                    end
+                end
+                push!(repl_code_chunks,
+                    line => String(strip(read(a, String))) * "\n"
+                )
             end
-            io = IOBuffer()
-            fun = getproperty(Pkg, Symbol(verb))
-            if isempty(args)
-                fun(; io)
-            else
-                fun(args; io)
-            end
-            push!(repl_code_chunks,
-                code => String(take!(io))
-            )
 
         elseif help
             # NOTE: this is pretty crap there should be a better way to just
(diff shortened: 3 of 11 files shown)

### CODE REVIEW

**Comment by gdalle** [comment]
This is amazing! Thank you SO MUCH 😍 Testing as we speak
Our main use case for the blog post was illustrating the pkg, help and cmd modes, is that also part of this PR?

**Comment by tlienart** [comment]
one thing that could be polished is to store code chunks and results in files so that we don't have to re-eval if nothing changes (at the moment these blocks will be re-evaluated on every page change).

**Comment by gdalle** [comment]
> one thing that could be polished is to store code chunks and results in files so that we don't have to re-eval if nothing changes (at the moment these blocks will be re-evaluated on every page change).

I spent hours this morning trying to combine Replay.jl with asciinema to show things like
```julia
pkg> st
```
or
```julia
help?> map
```
in a reproducible way, so honestly rerunning everything upon page changes is the last of my worries 🤣

tlienart:
> Our main use case for the blog post was illustrating the pkg, help and cmd modes, is that also part of this PR?

no unfortunately, these would potentially also be a fair bit more difficult to do (to give a hint: you can't copy paste in the REPL something with a `?` at the front and expect it to switch to help mode). 

We could imagine having a ```` ```? ````, ```` ```] ```` and ```` ```; ```` in a similar way as is done here but it'd require knowing how to run the relevant functions to retrieve the output of whatever the help mode calls etc. Definitely out of scope for the current PR. Your best option here is to just copy paste the REPL output into a `julia-repl` code block.

gdalle:
Thanks for the explanation!
I'll open an issue to keep track

gdalle:
The PR breaks on my macOS:
```julia
julia> using Franklin

julia> serve()
  Activating project at `~/Work/GitHub/ModernJuliaWorkflows`
→ Initial full pass...
┌ Franklin Warning: in <Project.toml>
│ Encountered an issue processing 'Project.toml' in ModernJuliaWorkflows.
│ Verify, then re-start the Franklin server.
│ The error is displayed below:
MethodError(Base._memcmp, (UInt8[0x5b, 0x64, 0x65, 0x70, 0x73, 0x5d, 0x0a, 0x46, 0x72, 0x61, 0x6e, 0x6b, 0x6c, 0x69, 0x6e, 0x20, 0x3d, 0x20, 0x22, 0x37, 0x31, 0x33, 0x63, 0x37, 0x35, 0x65, 0x66, 0x2d, 0x39, 0x66, 0x63, 0x39, 0x2d, 0x34, 0x62, 0x30, 0x35, 0x2d, 0x39, 0x34, 0x61, 0x39, 0x2d, 0x32, 0x31, 0x33, 0x33, 0x34, 0x30, 0x64, 0x61, 0x39,...
ERROR: MethodError: no method matching _memcmp(::Vector{UInt8}, ::Vector{UInt8}, ::Int64)

Closest candidates are:
  _memcmp(::Union{Ptr{UInt8}, AbstractString}, ::Union{Ptr{UInt8}, AbstractString}, ::Int64)
   @ Base strings/string.jl:129
```

tlienart:
This isn't related, it seems to be an issue with not being able to initiate a `newsite` which requires copying files and changing their access rights (long story short without doing this, a `newsite(...)` call would copy files in a read-only mode. Given that FranklinTemplates hasn't changed, I'm not sure why you're seeing this.

If you activate a project with master Franklin and another one with this branch, can you confirm that whatever steps you took work with the former and fail with the latter? thanks!

tlienart:
blank line should be removed

for the contrast, yes it's handled by highlight js; there are two things you could do here:

1. use a different style sheet (instead of the github one) using one of the style sheets here: https://highlightjs.org/static/demo/ 
2. manually customise the style sheet yourself

the style sheet are in `_libs/highlight/styles/` and to change the ones you use you'd have to do that in `_layout/head_highlight.html`.

tlienart:
@gdalle the basic stuff is added, you can test the following:

````
```?
sqrt
```
```]
st
```
```;
echo "foo"
```
````

**Note**: for the `help` mode, Julia returns stuff in Markdown format; it can convert it directly to HTML and so here this HTML is shown inside a div container `julia-help` which would have to be styled. I haven't really played with this yet. Maybe have a go and lmk what you think? showing just plaintext won't work because there'll be stuff like ```` ``` ```` and `[...](@ref)` shown.

**Comment by tlienart** [comment]
additions to the docs

Screenshot 2023-07-10 at 12 38 23

Screenshot 2023-07-10 at 12 38 18

**Comment by gdalle** [comment]
This is awesome, thank you so much! It will improve the feeling of our blog post by 100. Testing now

**Comment by gdalle** [comment]
For the REPL mode, here are my first impressions
- missing one line break between each prompt
- `nothing` outputs should not be printed
- errors are not shown?

image

**Comment by gdalle** [comment]
For the shell mode:
- the prompt only appears on the first line
- following lines are not interpreted
- errors are not shown

Test code

```
echo "hello"
echo "goodbye"
echography
```

Result

image

**Comment by gdalle** [comment]
For the Pkg mode:
- the prompt only appears on the first line
- the name of the environment is not shown
- adding a package fails
- doing anything after adding a package also fails

Test code

```
st
add Random
st
```

Results

image

**Comment by tlienart** [comment]
Thanks for the feedback @gdalle, some of these should be fixed in the current PR, some not. Notably capturing errors can be a bit annoying.

Separately: would you be open to moving to Xranklin for what you're building now? it might require a few adjustments but fixing some of the stuff you pointed out would be easier there (as I'm a bit reluctant to add a lot of code to Franklin `0.10`)

### stuff that can be done soon and easily

* [ ] repl - adding lines between `julia>` 
* [ ] repl - nothing handling
* [ ] shell - multi line, if we can make the assumption that each line is one `Cmd`
* [ ] pkg - multi line 

### stuff that is likely to be hard(er)

* [ ] help - multiline, likely too hard but I'm also a bit unsure of whether this is really an interesting use case
* [ ] repl - errors (I need to try first, maybe this isn't too bad)
* [ ] shell - errors (might be able to redirect `stderr`, would have to try first)
* [ ] pkg - there's some work here (already done in Xranklin) to have the page have its own environment, there's some finicky stuff to figure out for the add/remove etc. 
* [ ] multi line shell stuff if we cannot make the assumption that each line is a single command --> probably too hard, won't happen (there'd need to be something similar to the repl mode where you can figure out whether the AST is complete or not)

**Comment by gdalle** [comment]
> Thanks for the feedback @gdalle, some of these should be fixed in the current PR, some not. Notably capturing errors can be a bit annoying.

Maybe we don't have to capture errors very precisely, and as a first approximation we can just say that there's been an error without showing the full stack trace

**Comment by gdalle** [comment]
> Separately: would you be open to moving to Xranklin for what you're building now? it might require a few adjustments but fixing some of the stuff you pointed out would be easier there (as I'm a bit reluctant to add a lot of code to Franklin 0.10)

Sure, I can do that. In the long run I'd like to add these blog posts to the official Julia blog, which is why I chose Franklin, but I guess in the long run the official Julia blog will also pivot?

**Comment by gdalle** [comment]
>  help - multiline, likely too hard but I'm also a bit unsure of whether this is really an interesting use case

Agreed, I just tested it cause I thought it was an edge case, but as far as I'm concerned it's enough to just mention in the Franklin documentation that only single lines are allowed

**Comment by tlienart** [comment]
> guess in the long run the official Julia blog will also pivot?

yes, in fact I won't release `0.11` (what Xranklin will be) before I've made sure that the julialang website can use it easily

**Comment by gdalle** [comment]
>  pkg - there's some work here (already done in Xranklin) to have the page have its own environment, there's some finicky stuff to figure out for the add/remove etc.

I'm definitely gonna check it out. Ideally it would be cool if we could name the environments and reuse them from block to block. Cause right now by default the code executes in the website's main `Project.toml`, which is not ideal in terms of dependencies (I wonder how Julia blog posts handle this actually)

**Comment by gdalle** [comment]
So should I keep using this PR or switch to Xranklin (perhaps on a specific branch?)

**Comment by tlienart** [comment]
> (I wonder how Julia blog posts handle this actually)

I don't think any julia blog post uses evaluated code blocks

The `repl` items should all be fixed, including the error stuff (though it's not styled) so

````
```>
sqrt(-1)
```
````

for instance gives

```html
<pre><code class="language-julia-repl">julia> sqrt&#40;-1&#41;
DomainError with -1.0:
sqrt will only return a complex result if called with a complex argument. Try sqrt(Complex(x)).

</code></pre>
```

tlienart:
> So should I keep using this PR or switch to Xranklin (perhaps on a specific branch?)

Maybe keep using this PR for now, I'll have a look at whether I can backport the Pkg stuff without making too much of a mess. All this stuff is not in Xranklin yet (but it's an ok draft).

shashi:
@tlienart maybe it is easy to add an option to render the output using the richest MIME? Then we could have LaTeX output and Images!

Reminding you of
https://github.com/JuliaSymbolics/Symbolics.jl/pull/527 which I never completed :)

tlienart:
that's definitely out scope for the present PR, however on a more positive note, this is much improved in Xranklin which lets the user define their preferred "show" method for objects if you're interested see https://tlienart.github.io/Xranklin.jl/syntax/code_1/#output_of_executable_code_blocks. This already allows image stuff and should make LaTeX output a fair bit easier.

gdalle:
Indeed the Julia mode looks perfect now!
Still haven't managed to style documentation output though, so it looks like plain text. My changes to the CSS were ineffective, could you maybe take a look at https://github.com/gdalle/ModernJuliaWorkflows/tree/repl and tell me what you think?

tlienart:
@gdalle there's a stray `}` at the end of your `jtd.css` it should be

```css

.franklin-toc {
  /* Avoid clickable elements being too close together. */
  margin: 0.6rem 0;
}


/* ==================================================================
    REPL BLOCKS
================================================================== */

.julia-help {
  background-color: yellow;
  padding: 10px;
  font-style: italic;
}

.julia-help h1,h2,h3 {
  font-size: 1em;
  font-weight: 500;
}
```

gdalle:
If we want to get the right colors for the `pkg>` (blue), `shell>` (red) and `help?>` (yellow) prompts, does it happen in the Julia code or in the style sheets?

tlienart:
in `jtd.css` remove the `hljs.metap`, `hljs.metas` and replace them by 

```css
.hljs-meta.shell_ {color: red;}
.hljs-meta.pkg_ {color: cornflowerblue;}
```

tlienart:
yeah so if in `_libs/highlight/highlight.min.js` you add

```js
subLanguage:"julia"}},{className:"meta.help",begin:/^help\?>/,relevance:10,starts:{end:/^(?![ ]{6})/,
```

after 

```
subLanguage:"julia"}},{className:"meta.shell",begin:/^shell>/,relevance:10,starts:{end:/^(?![ ]{6})/,
```

and then in `jtd.css` add

```css
.hljs-meta.help_ {color: goldenrod;}
```

it'll get the styling right.

tlienart:
Pkg mode is added (including multiline). Each page can activate its own environment now (recommended is to activate a temporary one). The site-wide environment is re-activated at the end of the page by default. 

You can try

````
```]
activate --temp
add StableRNGs
```

then use it

```>
using StableRNGs
round(rand(StableRNG(1)), sigdigits=3)
```
````

gdalle:
Amazing! I'll test right away

gdalle:
That's really cool, the styling works (maybe we'll need a group PR to all Franklin templates?) and the multiline modes too.
Only trouble is with the Pkg environment activation shenanigans:
- When I don't do any `activate` in the code blocks, as soon as I interrupt the `serve()` process I land back in global environment `@v1.9` instead of my package environment `(MyBlog)`
- When I activate a temporary environment, it is retroactive and shows up in the previous prompts, possibly even in previous blocks (haven't tested). Is it just esthetic (ie only the printed name) or are all the code blocks live-reevaluated in the temporary environment?
image

tlienart:
both should be fixed now

gdalle:
This looks really good to me, I don't have any other requests (lucky you :rofl:).
We're gonna keep working on ModernJuliaWorkflows, in the short term how do you suggest we specify our dependencies? For now I depend on this specific `tl/fr818b` branch, do you plan on deleting it?

gdalle:
Two minor details:
- Spacing is a bit too narrow after the output of a help block
- Would it be possible for said output to have round corners like the other blocks in Franklin?

I fixed it in my style sheet but it might be worth harmonizing in the future

gdalle:
Sounds very reasonable, thank you again for your hard work! I'll make sure to mention you on the blog post

PR #1038, opened 2023-07-30
Replace html-css-js minify with minify_jll, eliminate Python3 dependency

M-PERSIC:
I recently ported over [Taco de Wolff's Minify ](https://github.com/tdewolff/minify.git) to Julia as a JLL package and felt it would be an excellent addition to Franklin.

### Advantages
- Dependency packaged directly as a JLL 
- Additional support for json, svg, and xml minifcation
- Eliminate Python3 as an external dependency for minification (I commented out the internal functions if they are ever needed again for some future dependency, though I can remove them outright if desired)

### Disadvantages
- Would need to bump Franklin to julia versions 1.6+. I could re-add the python3 minifer and ensure it is only used for julia v1.5 and below.
- Minor performance regression (2x based on some side-by-side comparisons) 

### Considerations
- I did not touch any GitHub actions for this pr as I'm unsure if we would still prefer installing python3 for whichever reason.


tlienart:
Hello @M-PERSIC , thanks a lot for this, it's awesome and the PR is nice and clean. I will not merge this into Franklin however as the current status is kind of like an LTS release. The next version of Franklin is in preparation at [Xranklin.jl](https://github.com/tlienart/Xranklin.jl/issues) and about 95% ready (mostly docs stuff at this point). 

The next version drops these external dependencies which caused issues and ultimately were not that useful. 

`minify_jll` sounds great but it's the kind of stuff which I'd like users to opt-in via their utils file rather than have it baked in the code.

Comments / next steps:

1. there isn't yet an explicit spot in the next version of Franklin to specify post-build logic; this is where a user who would want to benefit from `minify_jll` would add a few lines to indicate they want to do so. If you have thoughts on where this could be placed (e.g. in `utils.jl` or in a separate file, or as an argument to `build(...)`) and, more generally, if you're interested in porting this as a plugin to this next version, please kindly open an issue at Xranklin.jl for further discussion
2. in your code, you call `` success(`$(minify_jll.minify()) -o $(filepath) $(filepath)`) ``, I'm not very familiar with `jll` files but can't these be called directly within Julia? does it have to be shelled out? not an issue, just curiosity here.

Thanks again for your proposal!

M-PERSIC:
@tlienart Thank you for your feedback! 

- I think minification should be a fundamental component of Xranklin's build process as its advantages are too numerous. Some quick brainstorming, minify_jll could be included as a package extension and a minify option is added to the file watching process via `full_pass()`.  I'll open an issue in Xranklin hopefully soon once I have some more concrete ideas.
- It depends on what is included in the JLL package if it must be shelled out or not. If a C/Fortran-ABI compatible shared library is included, it can be invoked via Libdl or ccall. If an executable is included, then it must be run as an external program, though because it is fully packaged into Julia with the appropriate wrappers it is treated as a first-class Julia package. A good summary of JLL packages is in [Pkg + BinaryBuilder -- The Next Generation
](https://julialang.org/blog/2019/11/artifacts/#julia_library_jll_packages)

PR #1049, opened 2023-08-17
add sitemap_file as a global option

### LINKED ISSUE #1047: Option to set sitemap filename
We need to rename the `sitemap.xml` file for a Franklin site. We can just rename the file after the build, but it would be convenient if Franklin could do it for us. Would it be reasonable to have an option for this, e.g. `sitemap_filename`, that would default to `sitemap.xml`, but could be customized?

I'd be happy to look into a PR for this, but I could use some pointers to understand the codebase. I see it's currently hard-coded in these two places:

https://github.com/tlienart/Franklin.jl/blob/f1f7d044dc95ba0d9f368a3d1afc233eb58a59cf/src/manager/robots_generator.jl#L30

https://github.com/tlienart/Franklin.jl/blob/f1f7d044dc95ba0d9f368a3d1afc233eb58a59cf/src/manager/sitemap_generator.jl#L70

Would it be as easy as just adding a `globvar(:sitemap_file)` call to those places?

tlienart:
@mortenpi would you mind trying this out?

closes #1047 by adding a global variable `sitemap_file` if the user wants something else than `sitemap(.xml)`. 

**config.md**
```
+++
# ... rest
sitemap_file = "mySitemap"
# ... rest
+++
```

--> `mySitemap.xml` will be generated (instead of default `sitemap.xml`)

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "Franklin"
 uuid = "713c75ef-9fc9-4b05-94a9-213340da978e"
 authors = ["Thibaut Lienart <tlienart@me.com>"]
-version = "0.10.89"
+version = "0.10.90"
 
 [deps]
 Dates = "ade2ca70-3891-5945-98fb-dc099432e06a"
--- a/src/manager/robots_generator.jl
+++ b/src/manager/robots_generator.jl
@@ -26,8 +26,9 @@ function robots_generator()
     dst = joinpath(path(:site), "robots.txt")
     isfile(dst) && rm(dst)
     io = IOBuffer()
+    sitemap_file = splitext(globvar(:sitemap_file))[1] * ".xml"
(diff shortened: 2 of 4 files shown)


mortenpi:
Yep, this is perfect! Thank you!

Just as some bikeshedding: the option could also be called `sitemap_filename` maybe, to make it clear that it should not be a path?

tlienart:
Agreed, in the next version there are aliases for default variables which makes that stuff a bit easier (just add more aliases!) but agreed that filename is better, will add.

tlienart:
added `sitemap_filename` and `rss_filename` as aliases for `sitemap_file` and `rss_file`. will merge now, thanks for the FR!

mortenpi:
Thank you for implementing it!

PR #1091, opened 2024-09-02
fix verify_links to accept doi links

briochemc:
I'm not sure this is useful but many dois of mine get rejected and pop up during the link verification. Adding the x-bibtex header (which I spotted in https://github.com/JuliaWeb/HTTP.jl/issues/796) seems to fix that for me, hence this tiny PR. Do with it what you will!


briochemc:
PS: Thank you BTW! I made my website at www.bpasquier.com with Franklin, in case you want to add it to your list in the readme.

tlienart (approved):

tlienart:
sorry for the (very long) delay and thanks!!

PR #1105, opened 2025-10-21
Update a bunch of links

fingolfin:
There are still a bunch of references to https://github.com/tlienart2/tlienart2.github.io and https://github.com/tlienart2/myWebsite where I am not sure what to do with them.

Also, one link to http://github.com/tlienart/FranklinUtils.jl (note the `http` ;-) )


fingolfin:
I think CI fails because the bad URL is pulled from `FranklinTemplates.jl` -- while my fix for the URL was merged, it is not yet in a release of that package, and thus CI does not pick it up.

blegat:
Would be useful to merge this, it's quite confusing for a new user that the README links are broken

PR #1111, opened 2026-03-23
fix #909: consistent processing for HTML pages, normalize sitemap/RSS…

### LINKED ISSUE #909: inconsistency with add_rss_item, add_sitemap_item, add_disallow_item and on_write
Looking through the source code I stumbled upon different treatment, depending on whether it is markdown or html file, of the additions `add_rss_item`, `add_sitemap_item`, `add_disallow_item` and `on_write`.

[For the html case](https://github.com/tlienart/Franklin.jl/blob/c94317b5823c3510ed9180e7d374010c08557262/src/manager/file_utils.jl#L161-L163
) only `add_sitemap_item` is applied, and this explicitly BEFORE processing the html

[For the markdown case](https://github.com/tlienart/Franklin.jl/blob/c94317b5823c3510ed9180e7d374010c08557262/src/manager/write_page.jl#L200-L219) all four are applied, however all AFTER the html is processed

It would be nice to have this consistent

samtalki:
Fixes #909.

HTML pages ([`file_utils.jl:162`](https://github.com/JuliaDocs/Franklin.jl/blob/master/src/manager/file_utils.jl#L162)) previously only called `add_sitemap_item`. Markdown pages ([`write_page.jl:222-237`](https://github.com/JuliaDocs/Franklin.jl/blob/master/src/manager/write_page.jl#L222-L237)) called all four: `add_rss_item`, `add_sitemap_item`, `add_disallow_item`, and `on_write`.

HTML pages now call all four after conversion, matching the markdown path. `on_write` receives the rendered HTML string, consistent with the markdown path.

Also strips `index.html` from sitemap and RSS URLs by default (configurable via `sitemap_strip_index` in `config.md`) and adds depth-based sitemap priority as a sane default for non-markdown pages.


fkastner:
I like that this now strips `index.html` from the sitemap entries by default.
Is there anything left to do before this can be merged (and released)?

